PR #8, opened 2021-03-11
GPLv3+ → AGPLv3+ license

comradekingu:


rolandgeider:
Since flutter apps can be built for the web, we should change the license from GPL 3+ to AGPL 3+

Is this ok with you @stefanorossiti @Dolaned ?

stefanorossiti:
I dunno... honestly, before even considering the license itself, I checked this guy profile... 460+ branched repos, there are many pull requests "xxx to AGPLv3+" (the ones checked were refused, is interesting reading why). That's pretty... curious... to say the least. 
There are always pro and cons and if someone is only showing you the pros, he's not even trying to hide the fact he's just trying to convince you, not trying to find the best solution.

If you are concerned about the web version you could just add something specific for that at the actual license. 

Said that, I would advise you to read the license carefully and see all the differences. Don't forget this is something more legal than technical.

rolandgeider:
For reference, the original issue is https://github.com/wger-project/wger/issues/628

Both GPL and AGPL are pretty similar in that they are copyleft licenses. The main difference is that if your software was available via a network, you would not need to make the source available and AGPL closed this loophole, that's why the wger web application uses that license (obviously IANAL, but that's the gist). This would definitely be better than adding our own specific changes to the license, which would need to be legally correct and without unintended side effects.

stefanorossiti:
Yes ofc, just wanted to note that could be an option to forbid any right regarding the web build of flutter considering the situation.
But that's really down on what you want to do.
If you want the web side of flutter to be safely released and licensed and you are fine with AGPL, yes, switching to AGPL is something that should be done to prevent someone being able to close source his web only implementation.

rolandgeider:
See also #10

comradekingu:
There aren't many licenses suggested by me in those. You will find I for example my contribution to wger https://hosted.weblate.org/projects/wger/mobile/nb_NO/#history a better representation of what I do.
Did the same for Feeel.

stefanorossiti:
> There aren't many licenses suggested by me in those. You will find I for example my contribution to wger https://hosted.weblate.org/projects/wger/mobile/nb_NO/#history a better representation of what I do.
> Did the same for Feeel.

Rereading what I wrote I see that seem i little salty, sorry for that I was just in a rush and didn't really pay the due attention.
But I am still not sure AGPL would ideally be the best license for a flutter app. Not saying is bad tho.

comradekingu:
@stefanorossiti Just giving myself some representation. 
My original concern (which I should have stated) is AGPLv3 wger goes into GPLv3+ flutter repo how?
From my non-lawyer perspective an easy fix is to make them both AGPLv3+.

Dolaned:
im all good with this

stefanorossiti:
btw yes am fine too

rolandgeider:
> My original concern (which I should have stated) is AGPLv3 wger goes into GPLv3+ flutter repo how?
> From my non-lawyer perspective an easy fix is to make them both AGPLv3+.

Since there is no common code used and the only communication is via the REST API, there wouldn't be any license problems.

But that's not really important anymore, I'll merge this and #10 and update the license in weblate

PR #42, opened 2021-06-19
Added support for auto linting

### LINKED ISSUE #1: Use automatic linting
The code should be automatically linted for new commits or PRs. While github's super-linter supports dart, it seems it is not really usable at the moment:

https://github.com/github/super-linter/issues/507

M123-dev:
Closes #1 
This workflow runs on each push to the master branch, formats the code and commits the changes.

Alternatively, we can create a pr instead of a direct commit


rolandgeider:
Thanks!

PR #43, opened 2021-06-19
Fix for #42/ Automatic linting

M123-dev:
```line 1: syntax error near unexpected token `( ```

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -27,7 +27,7 @@ jobs:
       
       - name: Push a commit with the changed files
         run: |
-          git config user.name Github-actions (automatic linting)
+          git config user.name Github-actions
           git config user.email github-actions@github.com
           git add .
           git commit -m "Automatic linting"


M123-dev:
First run finished: https://github.com/wger-project/flutter/commit/7492b53c67597e21790487a83b50b3ccaba8b97f

rolandgeider:
I saw it! Having stuff like this automated is such a good feeling

rolandgeider:
Mhh, it seems it throws an error If there are no linting changes to commit

https://github.com/wger-project/flutter/runs/2869078524

M123-dev:
Thats not good, I just checked, doesnt look like github actions support any kind of allow-failure
https://github.com/actions/toolkit/issues/399, maybe we can check if anything has changed before running the commit step

rolandgeider:
Yes, perhaps it's possible to add some simple if checks in the job definition?

M123-dev:
Lets hope that https://github.com/wger-project/flutter/pull/47 fixes this problem

PR #44, opened 2021-06-19
CI fix locale zh_Hans

M123-dev:
Fixes CI:

```
Run flutter gen-l10n
Arb file for a fallback, zh, does not exist, even though 
the following locale(s) exist: [zh_Hans]. 
When locales specify a script code or country code, a 
base locale (without the script code or country code) should 
exist as the fallback. Please create a {fileName}_zh.arb 
file.
Error: Process completed with exit code 1.
```


M123-dev:
I just tested it and it is working for:

- zh_Hans_CN
- zh_Hans_HK
- zh_Hans_MO
- zh_Hans_SG

but nor for zh_Hans itself 🤔

rolandgeider:
Thanks!

Issue #38, opened 2021-06-12
Use correct weight for +/- buttons


rolandgeider:
In the gym mode, it's possible to add or substract weight by clicking on the plus / minus buttons, At the moment, the weight of the smallest plate is used (1,25), but it should be double (each side has one)

arun-muralidharan:
This should be a simple start. I'll work on this.

rolandgeider:
Closed with #45

PR #45, opened 2021-06-20
Plate Weight - Fix for #38

arun-muralidharan:
Fixed +/- button in gym mode to add twice the minimum plate weight, since we add one to each side.


rolandgeider:
Looks good!

Issue #35, opened 2021-06-11
Start time of workout session should be before the end


rolandgeider:
Add validator to enforce that the end and start times of a session (last page of gym mode) are in the correct order

rolandgeider:
I think TimeOfDay objects can't be directly compared, you will probably have to do a bit of data massaging

arun-muralidharan:
Yes, I'll have to create a DateTime object from TimeOfDay and then compare them using isAfter/isBefore. Something came up last night and I couldn't get to it. I'll do it this evening.

PR #49, opened 2021-06-22
Start Time Validator

arun-muralidharan:
Implemented start time validator to ensure start time is not ahead of end time in gym mode.

Also, fixed a bug where editing the end time changed the start time text.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -6,6 +6,7 @@
 * Roland Geider – https://github.com/rolandgeider
 * Dylan Aird - https://github.com/Dolaned
 * Jannik Norden - https://github.com/Jannik-dev
+* Arun Muralidharan - https://github.com/arun-muralidharan
 
 ## Translators
 
--- a/lib/helpers/misc.dart
+++ b/lib/helpers/misc.dart
@@ -16,6 +16,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import 'package:flutter/material.dart';
 import 'package:wger/helpers/consts.dart';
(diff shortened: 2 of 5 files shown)


arun-muralidharan:
Fix for [Issue#35](https://github.com/wger-project/flutter/issues/35)

rolandgeider:
Looks good, I'll test this later.

Can you add yourself to AUTHORS.md?

arun-muralidharan:
Will do, thanks!

rolandgeider:
The only thing I would perhaps change is adding a `errorMaxLines: 2` to the InputDecoration of the TextFormField where the new validator is, then the error message can be shown entirely.

And while this is not complex logic, I would suggest adding a simple test. I know writing tests is tedious and not fun, but I can't tell you how often they saved my ass while developing for wger. The good thing is that because the logic is simple, so are the tests! Something like this would have us covered:

```dart
void main() {
  test('Test the TimeOfDayExtension', () {
    final time1 = TimeOfDay(hour: 20, minute: 00);
    final time2 = TimeOfDay(hour: 10, minute: 00);
    final time3 = TimeOfDay(hour: 20, minute: 00);

    expect(time1.isAfter(time2), true);
    expect(time1.isBefore(time2), false);
    expect(time2.isAfter(time1), false);
    expect(time2.isBefore(time1), true);

    expect(time1.isAfter(time3), false);
    expect(time1.isBefore(time3), false);
  });
}
```

arun-muralidharan:
You are right. I'll make the edits. :)

arun-muralidharan:
I added the `errorMaxLines` and created a test file for extensions.

rolandgeider:
Great, thanks!!

Issue #41, opened 2021-06-16
Allow editing the exercise/set order


rolandgeider:
## Use case

it's currently not possible to change the order of sets once added to a workout.

## Proposal

There is already an order field for the sets that's populated during creation, so there is nothing that needs to be added on the backend. I'm not sure if we need a package for this or flutter has the necessary tools

arun-muralidharan:
There is Flutter's native `ReorderableListView` widget. We could use that; of if that doesn't work for us, there is also the [flutter_reorderable_list](https://pub.dev/packages/flutter_reorderable_list) package.

I'll start working on this.

rolandgeider:
Awesome! I'm not sure if the sets are currently rendered with ListTiles or something else, in case the  RecordableListView needs some specific child widgets to work

arun-muralidharan:
It is possible that `ReorderableListView` implements `ListView`. I'll take a look.

M123-dev:
@rolandgeider, is there even the possibility to do this on the webside?

rolandgeider:
There will be! It's on my to-do

PR #50, opened 2021-06-27
Reorder Sets

arun-muralidharan:
Progress PR for Issue #41.

Currently you can edit the order (long press to reorder) in the workout plan screen. The order change is only local; it persists only for the current session. I have not figured out how to reflect the changed order in the DB using the API. Does Day even store the order of its Sets in the DB? Or is it dynamically populated into the model in the app? I checked the model and its `toJSON()` ignores the Sets field. So, updating the Day using `editDay` has no effect.

### DIFF
--- a/lib/widgets/workouts/day.dart
+++ b/lib/widgets/workouts/day.dart
@@ -99,6 +99,7 @@ class _WorkoutDayWidgetState extends State<WorkoutDayWidget> {
   void initState() {
     super.initState();
     _sets = widget._day.sets;
+    _sets.sort((a, b) => a.order!.compareTo(b.order!));
   }
 
   void _toggleExpanded() {
@@ -135,7 +136,12 @@ class _WorkoutDayWidgetState extends State<WorkoutDayWidget> {
             visualDensity: VisualDensity.compact,
             icon: Icon(Icons.delete),
             iconSize: ICON_SIZE_SMALL,
-            onPressed: () {
+            onPressed: () async {
+              int _startIndex = _sets.indexOf(set);
+              setState(() {
+                _sets.remove(set);
+              });
+              _sets = await Provider.of<WorkoutPlansProvider>(context, listen: false).reorderSets(_sets, _startIndex);
               Provider.of<WorkoutPlansProvider>(context, listen: false).deleteSet(set);
             },
           ),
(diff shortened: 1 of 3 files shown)


rolandgeider:
Mhhh, I see. The first thing would be to make sure that the day's sets are always sorted by their order number. So basically we update the order field of the sets with each `onReorder` and because the provider sends out a `notifyListeners();` the view should get recreated on its own.

(I haven't tested this or anything, I'll try to take a look at the code this evening so I don't tell you any nonsense)

arun-muralidharan:
I think I get it now. I'll try to fix it today.

rolandgeider:
I see that there is no method to edit existing sets in the provider, but that's three lines, something like editDay

arun-muralidharan:
Yeah, just added that.

arun-muralidharan:
I made the changes. I added a `reorderSets` function and used it instead of `editSet` because multiple sets have to be edited even for one change as they are index-based. So, I didn't want to rebuild after each `editSet`, but after all of them are done. Also, even though Provider notifies and rebuilds after the change, I kept the local state because the UI change was smooth this way. Waiting for the notification made it lag otherwise.

rolandgeider:
you are right, it probably didn't update till it got the response from the server

lib/widgets/workouts/day.dart
```diff
@@ -137,6 +145,7 @@ class _WorkoutDayWidgetState extends State<WorkoutDayWidget> {
 
   @override
   Widget build(BuildContext context) {
+    _sets.sort((a, b) => a.order!.compareTo(b.order!));
```
rolandgeider, line 148: We would also use the sets in other parts of the application, it would be better to put that in the provider so that the correct order is used there as well. I think it's enough if we sort all the sets every time we add/delete one or call reorderSets

arun-muralidharan: I made the changes to addition and deletion sections. `reorderSets` loops and sets the order; I don't think we need to sort it there. Let me know if something is off.

rolandgeider:
Thanks for the PR!

I have added handles to the sets so that it's obvious that the order can be changed, a long press is not very discoverable. There is even an issue on flutter where they talk about this, the interface guidelines for material design use them and they are even rendered by default when building for the web, but for some reason, not for mobile.

PR #60, opened 2021-07-08
Bug: #48/ Fixed wrong workout name restriction

### LINKED ISSUE #48: Wrong workout name restriction (name length)
When renaming a workout or a workout day a minimum length of 5 letters is required. This limit does not exist on the server. (My workout as well as all of my workout days are under this limit.) Is this a wanted behavior?

M123-dev:
Fixes: #48 
And I also removed the pubspec.lock file because it is created in runtime and it can lead to errors if it comes from another system


rolandgeider:
are you sure about pubspeck.lock? The dart guide [recommends commiting](https://dart.dev/guides/libraries/private-files#pubspeclock) it (for applications)

M123-dev:
I had it in mind that it does not belong in the repo, no matter have undone the changes

PR #61, opened 2021-07-08
Bug #53/ double displayed training plans

### LINKED ISSUE #53: Double displayed training plans
When the app is loading (on startup) and you visit the Training section on the bottom navigation bar and you go back and revisit again it all workouts are displayed double(three times, four times, ....).

M123-dev:
Fixes #53 

Moved the FutureBuilder from dashboard.dart to home_tabs_screen.dart

At the moment the navigation bar is still shown, we can consider to also move it into the future builder and to show a nice splashscreen instead of the LinearProgressIndicator.

### DIFF
--- a/lib/screens/dashboard.dart
+++ b/lib/screens/dashboard.dart
@@ -38,88 +38,19 @@ class DashboardScreen extends StatefulWidget {
 }
 
 class _DashboardScreenState extends State<DashboardScreen> {
-  Future<void>? _initialData;
-
-  @override
-  void initState() {
-    super.initState();
-
-    // Loading data here, since the build method can be called more than once
-    _initialData = _loadEntries();
-  }
-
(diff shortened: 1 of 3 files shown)


rolandgeider:
Awesome, I'll take a look tomorrow!

rolandgeider:
this looks nice, but I'll try to remove the bottom navigation bar. Another thing we need to consider is when the list of workouts and nutritional plans get loaded/synced from the server. At the moment this happened when visiting the dashboard so that any new workout entered in the website also gets shown in the app

rolandgeider:
I've changed this so that the bottom navigation is not shown. Also, I managed to avoid doing this a bit longer than anticipated 😅

M123-dev:
Ahh I know, some things just stay undone, but it's always nice when they are finally done.

Issue #52, opened 2021-06-30
Automatically close the keyboard


M123-dev:
If you enter the weight or the repetitions via keyboard during a workout and go to the next exercise the keyboard remains open

PR #66, opened 2021-08-10
Automatically close keyboard in gym mode

KhushbuBora:
#52 
Added focusNode propert in gymMode to close keyboard when pressed on save button.

### DIFF
--- a/lib/widgets/workouts/gym_mode.dart
+++ b/lib/widgets/workouts/gym_mode.dart
@@ -259,11 +259,13 @@ class _LogPageState extends State<LogPage> {
   final _weightController = TextEditingController();
   var _detailed = false;
 
-  FocusNode focusNode = new FocusNode();
+  late FocusNode focusNode;
 
   @override
   void initState() {
     super.initState();
+
+    focusNode = FocusNode();
 
     if (widget._setting.reps != null) {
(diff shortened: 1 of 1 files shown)


M123-dev (approved):
The changes look good thanks.

-----------

Only one thing the line length is set 100 but the IDE's do not know that and format to the standard length. You can reformat by running: 
flutter format --line-length=100 .


@rolandgeider why is the line limit set to 100, I think it adds more effort than it brings code clarity

KhushbuBora:
> The changes look good thanks.
> 
> Only one thing the line length is set 100 but the IDE's do not know that and format to the standard length. You can reformat by running:
> flutter format --line-length=100 .
> 
> @rolandgeider why is the line limit set to 100, I think it adds more effort than it brings code clarity

Sure

rolandgeider:
I find 100 characters a good compromise, 80 seems too narrow to me. But we should mention this somewhere in the README or similar

rolandgeider:
I'm reading on the [flutter docs](https://flutter.dev/docs/cookbook/forms/focus) that the focus node should be created in initState and later disposed. Can you add these changes? Then this can be merged

KhushbuBora:
Okay..

Issue #72, opened 2021-08-28
User not notified about server side errors when updating weight. Results in seemingly unresponsive UI


patrickwitter:
## Steps to Reproduce

https://user-images.githubusercontent.com/53438425/131231052-0b224f23-9cd8-4380-b21e-9b088eb868a8.mp4

**Expected results:** 
A pop-up explaining to the user that a server error had occurred and some steps to resolve it. (if any)
Also, the user should be taken back to the dashboard or weight form
<br/>
**Actual results:** 
Unresponsive UI (from user's perspective) when save button is pressed

PR #74, opened 2021-08-30
Fixes Issue #72

patrickwitter:
Implemented Dialog Menu to show user server-side error. Fixes Issue #72 
Screenshot_1630269848

Removed cursor from pickDate Textfield and added icon. 
before:
Inkeddevice-2021-08-29-165603_LI
after:
Screenshot_1630269625


rolandgeider:
Thanks!

PR #87, opened 2021-10-01
Updated to use flutter_lints package

### LINKED ISSUE #86: Use flutter_lints package
From https://flutter.dev/docs/release/breaking-changes/flutter-lints-package  

> The package:flutter_lints defines the latest set of recommended lints that encourage good coding practices for Flutter apps, packages, and plugins. Projects created with flutter create using Flutter version 2.3.0-12.0.pre or newer are already enabled to use the latest set of recommended lints. Projects created prior to that version can upgrade to it with the instructions in this guide.

Since this project was created before that package was available, we need to migrate manually. Note that at the moment the app can't be build with flutter 2.5 due to https://github.com/google/charts/issues/678

M123-dev:
Fixes #86

The PR turned out a bit bigger than I would have hoped.

Since I had to intervene the lint settings a bit by hand I left out a few important things for now (marked with "# to be changed"), we can take care of those in future PR's. 

There were 3 true but small errors, which would have led to problems, I fixed them, otherwise there were no functional changes.

### DIFF
--- a/test/gallery_screen_test.mocks.dart
+++ b/test/gallery_screen_test.mocks.dart
@@ -105,8 +105,6 @@ class MockGalleryProvider extends _i1.Mock implements _i4.GalleryProvider {
       (super.noSuchMethod(Invocation.method(#deleteRequest, [url, id]),
           returnValue: Future<_i3.Response>.value(_FakeResponse_3())) as _i6.Future<_i3.Response>);
   @override
-  String toString() => super.toString();
-  @override
   void addListener(_i8.VoidCallback? listener) => super
       .noSuchMethod(Invocation.method(#addListener, [listener]), returnValueForMissingStub: null);
   @override
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -1,0 +1,2 @@
+include: package:flutter_lints/flutter.yaml
+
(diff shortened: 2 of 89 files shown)


rolandgeider:
wow, you weren't kidding! In any case, thanks a lot for taking the time, I'll merge this tomorrow

M123-dev:
I have to admit though, not everything was directly forced by flutter_lints, a few things also came through the linting rules

I just had a look why the tests fail, but couldn't reproduce it. Locally all the tests pass

rolandgeider:
it's always the same test, it sometimes fails on github

test/gallery_screen_test.mocks.dart
```diff
@@ -105,8 +105,6 @@ class MockGalleryProvider extends _i1.Mock implements _i4.GalleryProvider {
       (super.noSuchMethod(Invocation.method(#deleteRequest, [url, id]),
           returnValue: Future<_i3.Response>.value(_FakeResponse_3())) as _i6.Future<_i3.Response>);
   @override
-  String toString() => super.toString();
```
rolandgeider, line 108: FYI, all the foo.mocks.dart files are generated, so this will come back the next time we start the build runner

M123-dev: Ohh thanks I completely missed that, haven't really looked at the file names

PR #93, opened 2021-10-06
Correct Preset Time In Session Form

### LINKED ISSUE #31: Preselect correct time in session form
The session form (last page in the gym mode) always pre fills the current time when editing the start or the end time. It should of course use the values present in the field

TomerPacific:
# Proposed Changes
When a user chooses a start/end time at the end of the gym mode, the time will always be the same when trying to edit. This happened because of two reasons:

1. The initial values for the TimePicker were wrong
2. We were not setting the chosen value until the Save button was pressed

The fix adds a default time to session's timeStart/end and also corrects the behavior of setting these fields appropriately.

Fixes #31 

- [X] Set a 100 character limit in your editor/IDE to avoid white space diffs in the PR
- [X] Tests for the changes have been added (for bug fixes / features)
- [X] Added yourself to AUTHORS.rst

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -10,6 +10,7 @@
 * Khushbu Bora - <https://github.com/KhushbuBora>
 * Patrick Witter - <https://github.com/patrickwitter>
 * Tomer Ben-Rachel - <https://github.com/tomerpacific>
+* Thilina Herath - <https://github.com/ThilinaTCH>
 
 ## Translators
 
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -475,5 +475,9 @@
   "takePicture": "Take a picture",
   "chooseFromLibrary": "Choose from photo library",
   "gallery": "Gallery",
-  "addImage": "Add image"
+  "addImage": "Add image",
+  "dataCopied": "Data copied to new entry",
+  "@dataCopied": {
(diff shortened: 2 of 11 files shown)


lib/models/workouts/session.dart
```diff
@@ -42,10 +42,10 @@ class WorkoutSession {
   late String notes;
 
   @JsonKey(required: true, name: 'time_start', toJson: timeToString, fromJson: stringToTime)
-  late TimeOfDay timeStart;
+  late TimeOfDay timeStart = TimeOfDay.now();
```
rolandgeider, line 45: do we still need this?

rolandgeider: Since those are default values that are not constant, better move them to a new constructor, something like this, then we're good to go

```dart
WorkoutSession.now(){
    timeStart = TimeOfDay.now();
    timeEnd = TimeOfDay.now();
  }
```

TomerPacific: @rolandgeider, where would you like to call it so these fields will be initialized?

Just replace this line from gym_mode.dart:

`final _session = WorkoutSession();`
?

with this:

`final _session = WorkoutSession.now();`

rolandgeider: Yeah, exactly

TomerPacific: @rolandgeider, done.

Issue #82, opened 2021-09-16
Consistenly display nutritional values


rolandgeider:
At the moment, the nutritional values are displayed differently in many screens. Specially for places with little space we should use a compact and consistent way of displaying them.  The most obvious option would be to display the initials: E(nergy), C(arbs), P(rotein), F(at), but we need to take translations into account

sandimilohanic:
@rolandgeider Is it possible to take?

rolandgeider:
All yours

sandimilohanic:
@rolandgeider  These fields are a problem?

![Screenshot 2021-10-05 at 18 25 38](https://user-images.githubusercontent.com/7765448/136064776-e5cbc36c-c36c-4bb2-9259-d892b31f6d6d.png)

rolandgeider:
Yes, the bottom one (the table is big enough to show the full names), but also
* the log detail page (just click on one of the dates), the entries for the individual ingredients
* in the dashboard the values for the individual meals

rolandgeider:
The problem with the expanded list is that you can't compare the values (but I know that the current way is not perfect).  And now that I see, I would remove the sugars and saturated fats from the view, those would be only available on the detail page, but that's something you don't need to do.

I would leave the first table as it was, but use the initials on the list view and the dasboard

sandimilohanic:
All right, I'll add the initials to the list view and dashboard.

I will remove sugars and saturated fats from the list view, I do this in the same issue?

PR #94, opened 2021-10-07
Consistenly display nutritional values

sandimilohanic:
@rolandgeider I made changes. I noticeed that when saving changes IDE (VS code) it made automatic code correction, is that a problem?
If this is a problem, I will correct the code formatting.

Additional: Removed macronutrients sugars and saturated fats

#82 Consistenly display nutritional values


rolandgeider:
Looks good, thanks!

PR #98, opened 2021-10-07
Better linting rules 1

M123-dev:
For #87 cf: #86

I updated some further linting rules and updated the code accordingly, for an easier review process I've put the different types of changes into different commits (with a description of what the change was)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "dart.lineLength": 100
+}
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -1,52 +1,47 @@
 # Authors
 
-
 ## Developers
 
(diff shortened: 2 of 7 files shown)


M123-dev:
Now only four hints are in the analyzer:
Unbenannt

I dont know if the not used vars are planned to be used so I left them in and the other two are some flutter things which got deprectaed lately but I left them in for now since we are using a hybrid flutter version.

rolandgeider:
thanks again for another herculean PR ;)

the ones for the exercises will be removed, we're refactoring that part in another branch anyway. If the other one isn't used then they can be removed

M123-dev:
@rolandgeider 

I haven't understood you right, should I undo the changes in the exercises or will they just be overwritten

Also we had merge conflicts in the pubspec.lock file I fixed them manually, hopefully nothing broke

Issue #103, opened 2021-10-18
RiR slider doesn't work as it should


TaarnStar:
## Steps to Reproduce

1. Go to any workout and start it
2. On the exercise page check the option Set units and RiR
3. Try to change the RiR with slider

**Expected results:** the expectation is that the slider immediately shows the changes

**Actual results:** the slider doesn't move, after turning off the checkbox and later turning it on, the changes are visible on slider

![image](https://user-images.githubusercontent.com/17704741/137796352-a4637b22-6bfb-481e-94d8-fa9a094f69a5.png)
![image](https://user-images.githubusercontent.com/17704741/137796368-76ad2e40-b868-4ed4-b6d4-ee88e141264e.png)
![image](https://user-images.githubusercontent.com/17704741/137796603-30eb4c60-4875-4993-9a62-c9cf5e689246.png)

The error when clicking on slider is 

![image](https://user-images.githubusercontent.com/17704741/137796432-7e9fded2-94b4-4ce2-972f-3c97658fff09.png)

PR #113, opened 2021-10-29
Fix for RiR slider

TaarnStar:
# Proposed Changes

- When trying to move the slide on slider in RiR, there were initialization errors (#103)

The RiR slider value was a final variable so it couldn't be reassigned and reinitialized the value. I changed to use the final variable only on first initialization, for the rest of the slider using I made a temp variable that isn't final so everything works as it supposed to..

### DIFF
--- a/lib/widgets/workouts/forms.dart
+++ b/lib/widgets/workouts/forms.dart
@@ -771,6 +771,8 @@ class RiRInputWidget extends StatefulWidget {
 }
 
 class _RiRInputWidgetState extends State<RiRInputWidget> {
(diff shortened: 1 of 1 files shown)


rolandgeider:
You are right. I think I'll leave the warning for now, then we don't forget about this

TaarnStar:
I removed the new mutable variable and removed the final keyword for those two properties, the slider works fine now, only the warning is there of using a final keyword on an @immutable class

M123-dev:
Sorry @rolandgeider this error is probably on me😅

rolandgeider:
Don't worry, I also didn't notice. We should probably handle the objects differently and more cleanly

PR #123, opened 2022-01-11
Fix installing on Android 12

TheLastProject:
Hey there,

I noticed your F-Droid forum thread: https://forum.f-droid.org/t/problem-building-app-for-android-12/16158

After looking at [andrewc's reply](https://forum.f-droid.org/t/problem-building-app-for-android-12/16158/2?u=thelastproject) I decided to see what your AndroidManifest.xml looks like and it seems you don't have any exporting info: https://developer.android.com/about/versions/12/behavior-changes-12#exported

For future reference, always make sure to read the release notes and test your app (in an emulator for example) on the newest Android versions, **especially if you update targetSdk** (because that means you're telling Android your app is ready for new behavioural changes), so you can prevent such surprises in the future :)

Note that I haven't tested it, as I don't currently have all the stuff setup to build Flutter apps. So I'd suggest trying this patch, building a new .apk and testing it in an Android 12 emulator or asking @mondlicht-und-sterne to test it for you :)

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -30,6 +30,7 @@
 
         <activity
             android:name=".MainActivity"
+            android:exported="true"
             android:launchMode="singleTop"
             android:theme="@style/LaunchTheme"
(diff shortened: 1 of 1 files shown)


rolandgeider:
Hi! Thanks for the PR! The strange thing is that the version on the playstore that is built from the same source can be installed without problems on android 12.

In any case, I'll keep an eye on the release notes in the future 😅

PR #125, opened 2022-01-22
feat: Form Related Changes in Add Exercise Screen

agwanyaseen:
Form Changes
-

### DIFF
--- a/lib/providers/add_excercise_provider.dart
+++ b/lib/providers/add_excercise_provider.dart
@@ -2,9 +2,31 @@ import 'package:flutter/foundation.dart';
 
 import 'dart:io';
 
+import 'package:wger/models/exercises/muscle.dart';
+
 class AddExcerciseProvider with ChangeNotifier {
   List<File> get excerciseImages => [..._excerciseImages];
   final List<File> _excerciseImages = [];
+  String? _name;
+  String? _alternativeName;
+  String? _targetArea;
+  List<String?>? _primaryMuscles = [];
+  List<String?>? _secondaryMuscles = [];
+
+  set exerciseName(String name) => _name = name;
+  set alternateName(String? name) => _alternativeName = name;
+  set targetArea(String target) => _targetArea = target;
+  set primaryMuclses(List<String?>? muscles) {
+    if (muscles?.isNotEmpty ?? false) {
+      _primaryMuscles = muscles;
+    }
+  }
+
+  set secondayMuclses(List<String?>? muscles) {
+    if (muscles?.isNotEmpty ?? false) {
+      _secondaryMuscles = muscles;
+    }
+  }
 
   void addExcerciseImages(List<File> excercizes) {
     _excerciseImages.addAll(excercizes);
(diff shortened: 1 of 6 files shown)


lib/providers/add_excercise_provider.dart
```diff
@@ -2,9 +2,31 @@ import 'package:flutter/foundation.dart';
 
 import 'dart:io';
 
+import 'package:wger/models/exercises/muscle.dart';
+
 class AddExcerciseProvider with ChangeNotifier {
   List<File> get excerciseImages => [..._excerciseImages];
   final List<File> _excerciseImages = [];
+  String? _name;
+  String? _alternativeName;
```
rolandgeider, line 11: the alternative names should be a string since some exercises can have several names

lib/providers/add_excercise_provider.dart
```diff
@@ -2,9 +2,31 @@ import 'package:flutter/foundation.dart';
...
+  String? _targetArea;
+  List<String?>? _primaryMuscles = [];
```
rolandgeider, line 13: wouldn't it be better to store this a a list of Muscles? Same goes to the other stuff we get from the form, it will be easier for us later when we need to do work with those objects later

lib/screens/add_exercise_screen.dart
```diff
@@ -89,38 +105,51 @@ class _AddExerciseScreenState extends State<AddExerciseScreen> {
...
+    final muscles = exerciseProvider.muscles;
     return Form(
-      key: _basicStepFormKey,
+      key: formkey,
       child: Column(
         children: [
           AddExerciseTextArea(
             onChange: (value) => print(value),
-            title: AppLocalizations.of(context).name,
+            title: '${AppLocalizations.of(context).name}*',
```
rolandgeider, line 123: I spoke with the guy that made the designs, we'll change a little bit on the part where the user enters the name and description. I'll write you later

lib/providers/add_excercise_provider.dart
agwanyaseen: Currently DropDowns and MultiSelect widgets are designed to handle only Strings and not Object, that is the reason I kept it as string.

lib/providers/add_excercise_provider.dart
agwanyaseen: It's already string!

lib/providers/add_excercise_provider.dart
rolandgeider: Ah, ok. Can we store the object IDs and do a lookup when saving?

lib/providers/add_excercise_provider.dart
rolandgeider: Sorry, I meant a list of strings

lib/providers/add_excercise_provider.dart
agwanyaseen: that's possible , but then in UI we will only see ID and not name.
Instead we need to re-design the widgets.

rolandgeider: Showing only the IDs is not optimal 😄, I guess we will have to think of something

lib/providers/add_excercise_provider.dart
agwanyaseen: Every string seperated by space is new name. Correct?

rolandgeider: Either space or newline, yes. I think newline is more obvious since some names can have whitespaces in them

lib/providers/add_excercise_provider.dart
agwanyaseen: Cool then we need to re-design the widgets,
I'll try to do them.

lib/providers/add_excercise_provider.dart
agwanyaseen: I need to think,  how we can implement these.

agwanyaseen: These part is still pending, but branch is merged?

PR #132, opened 2022-03-13
Added translation into Japanese 

WMs784:
# Proposed Changes

- Added translation into Japanese on lib/l10n/app_ja.arb with reference to app_en.arb

I'm a Japanese native speaker.
I didn't complete translation of all words yet, but I’ll keep  translating if this PR is working.

- [lib/l10n/app_en.arb](https://github.com/wger-project/flutter/blob/master/lib/l10n/app_en.arb) L298: "Start time cannot be ahead of end time” is maybe mistake of “End time cannot be ahead of start time"

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -46,3 +46,7 @@
 - Norwegian Bokmål
 
   - Allan Nordhøy <epost@anotheragency.no> (98)
+  
+- Japanese
+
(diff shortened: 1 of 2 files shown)


rolandgeider:
Thanks for the PR!!

You are right, that string should be "End time cannot be ahead of start time", I'll change that

PR #192, opened 2022-06-25
Fix sub-url handling

kovdan01:
Previously the app was unable to work with instances installed on sub-urls
because sub-url path was removed (say, server.name/api/v2/login instead of
server.name/wger/api/v2/login).

Related Issues: https://github.com/wger-project/flutter/issues/193

### DIFF
--- a/lib/providers/helpers.dart
+++ b/lib/providers/helpers.dart
@@ -26,7 +26,7 @@ Uri makeUri(
 ]) {
   final Uri uriServer = Uri.parse(serverUrl);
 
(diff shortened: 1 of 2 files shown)


rolandgeider:
The idea with `uriServer.path` works and should solve your issue with the subpath, but there are probably some edge cases we need to consider (e.g. trailing slash/no trailing slash for the server), but it shouldn't be much. In any case I'll use this PR as a base

rolandgeider:
I just came back from vacation, will try to go over the open prs these next days

PR #237, opened 2022-09-30
Removing Bug which allows users to enter whitespace which is not accepted as a valid username

JustinBenito:
…pted.

Users mistakenly enter space which strikes an error, to stop this, entering space is ignored and not included, instead of throwing an error.

# Proposed Changes

-
-
-

-

### DIFF
--- a/lib/screens/auth_screen.dart
+++ b/lib/screens/auth_screen.dart
@@ -221,6 +221,7 @@ class _AuthCardState extends State<AuthCard> {
                       }
                       return null;
                     },
+                    inputFormatters: [FilteringTextInputFormatter.deny(new RegExp(r"\s\b|\b\s"))],
                     onSaved: (value) {
                       _authData['username'] = value!;
                     },


lib/screens/auth_screen.dart
```diff
@@ -221,6 +221,7 @@ class _AuthCardState extends State<AuthCard> {
                       }
                       return null;
                     },
+                    inputFormatters: [FilteringTextInputFormatter.deny(new RegExp(r"\s\b|\b\s"))],
                     onSaved: (value) {
                       _authData['username'] = value!;
```
rolandgeider, line 226: Perhaps it would be better to just trim the whitespaces when saving to _authdata?

JustinBenito: Yes It would be better , but What happens is when we enter a whitespace the regexp catches it and throws an error, Normal People will be frustrated by seeing an Error, It spoils our User Experience, So what we do as a developer is we ignore those mistakes made by the users, That is what most important in UX, So inorder to improve our UX(User Experience) we must ignore the mistake and not throw an error.

rolandgeider: yeah I meant instead of the regex, there might be some corner cases where the user only enters white spaces or somthing like that but we can probably live with that

JustinBenito: So this is a valid PR and I get a hactober-accepted label ;)

PR #238, opened 2022-10-01
These are the UI changes

JustinBenito:
# Proposed Changes

Enhanced UI/UX

## Please check that the PR fulfills these requirements

- [ ] Set a 100 character limit in your editor/IDE to avoid white space diffs in the PR
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Added yourself to AUTHORS.md

PR #246, opened 2022-10-02
Dark mode

JustinBenito:
I have now added the facility to add dark mode,
Light mode is now available seperately. So if you wish to have dark mode you may need to change only one line of code.

Related Issues:

Yes there is one issue that is, though I have made everything seemless to switch, the darkmode button is not implemented, so once you merge this, you may need to add a button which just changes true or false in the theme.dart file

I have only made the Dark mode available to develop more on, The theme is ready but the button to change is not. 
/* This is the Dark mode Theme */


rolandgeider:
I have moved the files to the correct locations as far as I could tell, some of them where duplicated.

Also, instead of doing `darkmode ? foo : bar` in the code, we should do something like https://stackoverflow.com/questions/60232070/how-to-implement-dark-mode-and-light-mode-in-flutter so that flutter takes care of all of that

```dart
MaterialApp(
      title: 'App Title',
      theme: ThemeData(
        brightness: Brightness.light,
        /* light theme settings */
      ),
      darkTheme: ThemeData(
        brightness: Brightness.dark,
        /* dark theme settings */
      ),
      themeMode: ThemeMode.dark, 
      /* ThemeMode.system to follow system theme, 
         ThemeMode.light for light theme, 
         ThemeMode.dark for dark theme
      */
      debugShowCheckedModeBanner: false,
      home: YourAppHomepage(),
    );
```

Plus here are the changes from #238 as well :) And here instead of doing 

```dart
Padding(
      padding: const EdgeInsets.all(8.0),
      child: Card(
        shape: RoundedRectangleBorder(
          borderRadius: BorderRadius.circular(15.0),
        ),
       .....
```
we should add a new widget and simply pass the child as an argument, then we don't duplicate so much code

rileyhawk1417:
@rolandgeider just to ask have these changes been made to the PR yet?

rolandgeider:
No, don't think so

rolandgeider:
Some of the changes here have been manually moved to the feature/material3 branch, closing here

PR #250, opened 2022-10-05
addded measurements to dashboard, categories card refactor

bujdy:
Added measurements to dashboard
Categories card refactor


lib/widgets/dashboard/widgets.dart
```diff
@@ -35,14 +36,20 @@ import 'package:wger/screens/workout_plan_screen.dart';
 import 'package:wger/theme/theme.dart';
 import 'package:wger/widgets/core/charts.dart';
 import 'package:wger/widgets/core/core.dart';
+import 'package:wger/widgets/measurements/categories_card.dart';
 import 'package:wger/widgets/nutrition/charts.dart';
 import 'package:wger/widgets/nutrition/forms.dart';
 import 'package:wger/widgets/weight/forms.dart';
 import 'package:wger/widgets/workouts/forms.dart';
 
+import '../../providers/measurement.dart';
```
rolandgeider, line 45: Can you use absolute imports? then this looks a bit more consistent

lib/widgets/dashboard/widgets.dart
```diff
@@ -96,7 +104,8 @@ class _DashboardNutritionWidgetState extends State<DashboardNutritionWidget> {
               icon: const Icon(Icons.history_edu),
               color: wgerPrimaryButtonColor,
               onPressed: () {
-                Provider.of<NutritionPlansProvider>(context, listen: false).logMealToDiary(meal);
```
rolandgeider, line 99: Could you set the line length to 100 characters? Then we can avoid these diffs

lib/widgets/dashboard/widgets.dart
```diff
@@ -309,6 +327,101 @@ class _DashboardWeightWidgetState extends State<DashboardWeightWidget> {
...
+  @override
+  Widget build(BuildContext context) {
+    final _provider = Provider.of<MeasurementProvider>(context, listen: false);
+
+    var items = _provider.categories
+        .map(
+          (item) => CategoriesCard(item),
+        )
+        .toList();
+    return Consumer<BodyWeightProvider>(
```
rolandgeider, line 350: Shouldn't this be the MeasurementProvider?

lib/widgets/dashboard/widgets.dart
```diff
@@ -309,6 +327,101 @@ class _DashboardWeightWidgetState extends State<DashboardWeightWidget> {
...
+                          );
+                        }).toList(),
+                      ),
+                    ),
+                  ])
+                else
+                  NothingFound(
+                    AppLocalizations.of(context).noWeightEntries,
+                    AppLocalizations.of(context).newEntry,
+                    WeightForm(),
```
rolandgeider, line 414: I'm not sure what to do here if there are no categories. From one side we could simply allow the user to add one, but it is then probably a bit strange that they can't add another one. It might be enough to just display some info text and add a button that takes the user to the measurement overview page

rolandgeider (changes requested):

lib/widgets/dashboard/widgets.dart
bujdy: Maybe there can be arrow -> (alligned on the right from title for example "Weights" or "Measurements") and on press function will be navigate to "Weights" or "Measurements" screens.

bujdy: something like this... 
image

also weights tab should be changed as measurements (add icon wont be there but at same row as "go to detail page"
image

rolandgeider: yes, in any case all cards should behave the same with respect to going to the detail page or adding entries

rolandgeider (approved):

rolandgeider:
Merged, thanks!

Issue #297, opened 2022-11-26
Instead of showing exercise names the app shows exercise id


mohammadrafigh:
## Steps to Reproduce

![signal-2022-11-22-20-43-46-125](https://user-images.githubusercontent.com/6737508/204082450-673a26aa-32c1-4525-bff8-80e4bf375bc6.jpg)

1. Start training mode
2. Press menu button on top right of the screen
3. You see ids instead of names

**Expected results:**

Get exercise names.

**Actual results:**

We get ids.

Harshit933:
hey! can you please assign me to this issue

rolandgeider:
Sure, go ahead. You can also take a look at the PR #300 which has already some work put into it

Harshit933:
On it!

aman123424:
hey! Can you please assign this issue to me, if not solved yet then

rolandgeider:
Hi @aman123424 ! I think @Harshit933 is still working on this

aman123424:
Okay, @Harshit933 , can we collaborate if you need help to resolve this?

Harshit933:
Hey you can work on this issue, I am unassigning myself.

aman123424:
Okay sure

jagrutatwork:
Is this issue resolved?

PR #300, opened 2022-11-26
Show exercise names in gym mode navigation

mohammadrafigh:
# Proposed Changes

- Fetch exercise names from provider instead of showing ids in gym mode navigation dialog

- #297 

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -17,6 +17,7 @@
 - Florian Schmitz - <https://github.com/floodoo>
 - Aman Negi - <https://github.com/AmanNegi>
 - Sandi Milohanic - <https://github.com/sandimilohanic>
+- Mohammad Rafigh - <https://github.com/mohammadrafigh>
 
 ## Translators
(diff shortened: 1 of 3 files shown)


rolandgeider:
Now that I'm looking at the code. Wouldn't it be easier if we modified `_exercisePages`? Right now it is basically only a list that contains the UUID and the position. We could convert it e.g. to a `Map<ExerciseBase, int>` and this would allow us to even show the exercise image in the dropdown (don't know if this would be a good idea UX wise, but we would have the possibility)

mohammadrafigh:
@rolandgeider yes sure, I did this first, but I got some exceptions due to unknown Map types. I know I should somewhere define a fromJson for that type to work but since I'm a newbie on Dart so avoided that :D. I will take a look again and try to go that way.

rolandgeider:
we don't need to convert to json, we should be able to just put the whole object there. But if this makes problems, we can just use your approach

Issue #258, opened 2022-10-12
Block user from accessing exercise contribution page


rolandgeider:
If a user has not confirmed their email (user provider, Profile), they should't be able to access the page to contribute exercises (AddExerciseScreen, Workout tab -> three dot menu). Instead, we should show a simple info text and a button that opens the user pofile form (UserProfileForm) where it is possible to confirm it

Hirdey-1999:
Can i work on this issue as i have did the same in my app.

rolandgeider:
Hi! Sure, it's yours

Hirdey-1999:
Thanks for assigning

Arnav2001:
can I work on this issue, I can resolve it.

rolandgeider:
Hey! @Hirdey-1999 is already working on this

Arnav2001:
do you have anything else for me

rolandgeider:
@Arnav2001 perhaps #254 ?

Harshit933:
Hey can you please assign me on this issue?

rolandgeider:
Hi @Harshit933 , sure!

If you need help setting up a backend or want to use the test server, just ping me

Harshit933:
@rolandgeider when I am running pub get i am getting this error, error is thrown from the l10n folder

![Screenshot from 2023-01-27 19-14-37](https://user-images.githubusercontent.com/90508384/215102840-7375faef-def6-49b6-969c-f356206fdf70.png)

I am quite new in open source any help would be much appreciated.

rolandgeider:
it seems there's something wrong with the translation files, let me take a look

Harshit933:
Okay, let me know whether should I file this as an issue.

rolandgeider:
It seems the placeholders in the translations were translated as well, if you pull from master it should work

rolandgeider:
BTW, you can use this server, just click on custom server in the login screen

https://wger-master.rge.uber.space/
username: user
password: flutteruser

(this is reset every 24h so feel free to change things)

Harshit933:
@rolandgeider There are still some issues. 
![Screenshot from 2023-01-27 21-32-00](https://user-images.githubusercontent.com/90508384/215133837-89be98f0-568f-46e6-bd3e-251eb96b520d.png)

rolandgeider:
fixed, now the app starts 😇 

Just had corrected those three errors from the first screenshot

Harshit933:
[...]

PR #332, opened 2023-01-27
User can't be able to add new exercise if their email is not verified

Harshit933:
# Added screen which redirects to the user form whenever the user tries to add an exercise when their email is not verified.

-
-
-

-


Harshit933:
fixes #258

rolandgeider:
Great, will try to review this later today

lib/screens/add_exercise_screen.dart
```diff
@@ -66,56 +75,115 @@ class _AddExerciseScreenState extends State<AddExerciseScreen> {
...
+    return Scaffold(
+      appBar: EmptyAppBar('Verify Email'),
+      body: Container(
+        padding: EdgeInsets.all(25),
+        child: Center(
+          child: Column(
+            mainAxisAlignment: MainAxisAlignment.center,
+            children: [
+              Text(
+                'Your profile needs to be verified in order to add any exercise',
```
rolandgeider, line 149: Can you mark the text as translatable? You just need to add an entry to app_en.arb and then you can use it with `AppLocalizations.of(context).foo` (sometimes the IDE doesn't pick this up right away and complains that foo doesn't exists, but the app builds just fine).

I would also suggest that we use the same text as in the web application: "You can only contribute exercises if your account is older than {{days}} days and have verified your email". We can just store the number of days in consts.dart with 14, at the moment this is not configurable.

lib/screens/add_exercise_screen.dart
```diff
@@ -66,56 +75,115 @@ class _AddExerciseScreenState extends State<AddExerciseScreen> {
...
+                    arguments: FormScreenArguments(
+                      AppLocalizations.of(context).userProfile,
+                      UserProfileForm(_user!),
+                    ),
+                  );
+                },
+                child: Container(
+                  padding: EdgeInsets.all(15),
+                  child: Text(
+                    'Profile Screen',
```
rolandgeider, line 174: Same here, just use `AppLocalizations.of(context).userProfile` (and something like a TextButton?)

Harshit933:
Got it!

Harshit933:
@rolandgeider done all the fixes!

rolandgeider:
Merged, thanks

PR #336, opened 2023-02-01
Update README.md

jakeau:
Updated readme to reflect flutter version

Updated project to enable building and deploying onto app store.


rolandgeider:
Is there a reason why so many files have been touched?

jakeau:
Had issues with the current pod setup for IOS so reset that. Also, the IOS verification process required app images to have no level of alpha in them which explains all the Asset file changes,  

Seems that a lot of the files are identical to the previous (fastlane + integration_test) can have a look at them and revert them after work today, Pushed this just before I went to sleep and didn't pay much attention to this all. Did ensure that it still worked before I pushed it.

rolandgeider:
ah, alright. At least the muscle SVGs need the alpha channel, we stack them to show the trained muscles for an exercises or workout. And the logo should also have transparency

Issue #293, opened 2022-11-17
Don't use canLaunchUrl to check whether a URL can be launched


12people:
From the [URL Launcher](https://pub.dev/packages/url_launcher) documentation:

```
If you need to know at runtime whether a scheme is guaranteed to work before using it (for instance, to adjust your UI based on what is available), you can check with [canLaunchUrl](https://pub.dev/documentation/url_launcher/latest/url_launcher/canLaunchUrl.html).

However, canLaunchUrl can return false even if launchUrl would work in some circumstances (in web applications, on mobile without the necessary configuration as described above, etc.), so in cases where you can provide fallback behavior it is better to use launchUrl directly and handle failure. For example, a UI button that would have sent feedback email using a mailto URL might instead open a web-based feedback form using an https URL on failure, rather than disabling the button if canLaunchUrl returns false for mailto.
```

This leads to links not opening inside e.g. Linux Flatpaks.

I recommend using this approach instead:
```
static void launchURL(BuildContext context, String url) async {
  final scaffoldMessenger = ScaffoldMessenger.of(context);
  final launched = await launchUrl(Uri.parse(url));
  if (!launched) {
    scaffoldMessenger.showSnackBar(SnackBar(
      content: Text("Could not open URL.".i18n),
    ));
  }
}
```

Issue #213, opened 2022-08-08
Build and publish the app as a snap/ flatpak


rolandgeider:
(Low prio issue)

The app should build without much effort as a linux desktop application (we probably want to work a bit on #111 before). We could think about publishing on [snapcraft](https://snapcraft.io/build) and [flatpak](https://docs.flatpak.org/en/latest/) if all plugins work there.

12people:
I can help with the Flatpak. (Not a fan of Snap, due to its server side being proprietary.)

Here's a manifest for a Flutter Flatpak I released that might serve as inspiration: https://github.com/flathub/com.enjoyingfoss.Parlera

rolandgeider:
yes, that would be great (but after we release the exercise stuff 😅 )

Ombrelin:
I too would love a release of the app for the desktop via flatpak !

rolandgeider:
Some TODOs

- [ ] adding desktop support doesn't currently work out of the box https://github.com/flutter/flutter/issues/53106
- [ ] uploading pictures to the gallery didn't seem to work

rolandgeider:
pull request sent https://github.com/flathub/flathub/pull/4075

PR #367, opened 2023-03-22
Linux/macOS support

12people:
# Proposed Changes

- Adding Linux support
- Adding initial macOS support
- Adding a Flatpak generator for future Flatpak releases

- #293 
- #213


lib/screens/gallery_screen.dart
```diff
@@ -35,20 +37,22 @@ class GalleryScreen extends StatelessWidget {
...
-        child: const Icon(Icons.add),
-        onPressed: () {
-          Navigator.pushNamed(
-            context,
-            FormScreen.routeName,
-            arguments: FormScreenArguments(
-              AppLocalizations.of(context).addImage,
-              ImageForm(),
-              hasListView: true,
+      floatingActionButton: (Platform.isLinux || Platform.isMacOS)
```
rolandgeider, line 40: one last question, why do you hide the floating action buttons on the desktop? Currently (at least on a couple of screens) this is the only way to add new entries. For other checks (like showing videos or using the scanner e.g.) I understand that they are still problematic on those platforms

12people: I believe this one is because `image_picker` is not available for Linux or for macOS, and the button leads to it.

It wouldn't make sense to have a button lead nowhere, and rewriting this functionality to work with a different library that does support Linux and macOS would require some time, which I don't have right now.

rolandgeider: ahh, the image picker, hadn't thought of that, you're right

rolandgeider:
Merged! Thanks :)

I'll see that everything flatpak/flathub related works and is submitted

Issue #306, opened 2022-11-28
Gallery Screen


jfrys:
## Use case

The Gallery Screen is in need of a prompt for users.

<img width="337" alt="gallery" src="https://user-images.githubusercontent.com/93684643/204167669-dbccf3c8-8436-44db-88ef-a6362564f2a7.png">

## Proposal

Since the Gallery is included on the main tab bar - the initial screen should have some sort of prompt for users to know what kind of media they should be uploading to this screen. I am not too sure personally what pictures would be best here, whether it's progress photos of myself, the food I eat or anything else that could relate to this app. I believe a simple text prompt of what wger is looking for users to upload on this screen would go a long way.

tavkeer:
can i contribute to this issue

tavkeer:
> ## Use case
> The Gallery Screen is in need of a prompt for users.
> 
> <img alt="gallery" width="337" src="https://user-images.githubusercontent.com/93684643/204167669-dbccf3c8-8436-44db-88ef-a6362564f2a7.png">
> 
> ## Proposal
> Since the Gallery is included on the main tab bar - the initial screen should have some sort of prompt for users to know what kind of media they should be uploading to this screen. I am not too sure personally what pictures would be best here, whether it's progress photos of myself, the food I eat or anything else that could relate to this app. I believe a simple text prompt of what wger is looking for users to upload on this screen would go a long way.

does this issue still needs work

tavkeer:
can i

tavkeer:
> ## Use case
> The Gallery Screen is in need of a prompt for users.
> 
> <img alt="gallery" width="337" src="https://user-images.githubusercontent.com/93684643/204167669-dbccf3c8-8436-44db-88ef-a6362564f2a7.png">
> 
> ## Proposal
> Since the Gallery is included on the main tab bar - the initial screen should have some sort of prompt for users to know what kind of media they should be uploading to this screen. I am not too sure personally what pictures would be best here, whether it's progress photos of myself, the food I eat or anything else that could relate to this app. I believe a simple text prompt of what wger is looking for users to upload on this screen would go a long way.

can i work on this

PR #378, opened 2023-04-17
[wger-project/flutter#306] Create text prompt for empty screens

freddyc2003:
# Proposed Changes

- Text displayed to the user when screens are empty prompting them to use a particular feature of the app,
- wording is generic across the Workout plans, Nutritional plans and Gallery screens to facilitate easy translation.

### DIFF
--- a/lib/widgets/core/text_prompt.dart
+++ b/lib/widgets/core/text_prompt.dart
@@ -1,4 +1,5 @@
 import 'package:flutter/material.dart';
+import 'package:flutter_gen/gen_l10n/app_localizations.dart';
 
 class TextPrompt extends StatelessWidget {
   @override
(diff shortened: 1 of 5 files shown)


lib/widgets/core/text_prompt.dart
```diff
@@ -0,0 +1,22 @@
...
+
+class TextPrompt extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    return Center(
+      child: Column(
+        mainAxisAlignment: MainAxisAlignment.center,
+        children: [
+          Text(
+            'Ready to start?',
```
rolandgeider, line 11: Can you mark these strings as translatable? Just add them to app_en.arb and then you can access the key with `AppLocalizations.of(context).yourKey`

rolandgeider: (the generated files are created on every build but sometimes the IDE still can't recognize them and mark the AppLocalizations key, but if it builds, just ignore the error)

freddyc2003: I have marked the strings as translatable and commited the changes.

rolandgeider: thanks, looks good! Will run some tests later today and then we can merge this :)

rolandgeider:
Merged, thanks!

PR #442, opened 2023-10-26
Add support for themed icon on Android 13+

0scvr:

Since Android 13, users can enable themed icons on their home screen.
Example:
image

- Added themed icon support for devices running Android 13 and above.
-

## Link to the issue : 

None.

## Tests

Not needed.

## Checklist


0scvr:
@rolandgeider What do you think ?

rolandgeider:
This is part of the new material me right? We still need to do some other work that is blocking us from moving ahead with material 3 (and dark mode), but sure, this could definitely be an option

rolandgeider:
So it seems we will need to do a bit more, flutter_launcher_icons doesn't seem to automatically generate the correct icons (and it seems it's half unmaintained?). We will probably need to generate the *_monochrome versions with some other tool

rolandgeider:
This was added with @12people 's PR, which also updated the icons to vector graphics, so closing here (but thanks anyway!)


Screenshot_20231109_203440

Issue #433, opened 2023-10-10
Chart polishing


rolandgeider:
List of things to do (on the `fix/replace-chart-lib` branch)

All
- [x] Check if spacing of values in y axis breaks depending on value ranges. Ideally we'd have two labels besides the start and end ones
- [x] Better spacing of x axis. It seems that fl charts always shows entries at the beginning and the end of the axis. Since this is not some important financial chart and should just show a tendency, I'd say that one or two more labels are enough
- [ ] Better Date format, MMMd or yMd if there are entries from last year

Workout log (`LogChartWidgetFl`)

- [x] Better margins
- [x] Fix chart colors (fix helper function so that we always get the same color from the same rep range). Note that we need to draw a line for every repetition range the user has done (so e.g. all entries for 8 reps, all for 10, all for 12, etc.) Realistically there probably isn't going to be too many, but we need to make sure everything still works fine
- [x] Add legend below the axis. Each color corresponds to a repetition range (e.g. blue for 8 reps, orange for 10 reps)
- [x] Fix tooltip. Ideally we would do something like "🟦 8 reps: 51kg"
- [ ] The labels for the y axis looks ugly and too far away from the axis but we need to consider that we can have five digits in total plus the weight unit (which is currently only "kg"). If there is an easy way to dinamically change this, we could try it+

![Screenshot_20231010_173428](https://github.com/wger-project/flutter/assets/2280729/529a015d-f063-4ed9-896f-ff0ef13c3ed3)

Nutrition diary chart (`NutritionalDiaryChartWidgetFl`)

- [x] Add a legend (alternatively, we could incorporate it to the heading)

![Screenshot_20231010_174336](https://github.com/wger-project/flutter/assets/2280729/836c7490-ecdc-4872-ba37-255ecff2a44e)

Measurement chart (`MeasurementChartWidgetFl`)

- [x] Move from core/charts.dart to measurement
- [x] Add the unit to the tooltip (so e.g. "80 kg" or "12 %")

![Screenshot_20231010_174506](https://github.com/wger-project/flutter/assets/2280729/19954485-d1eb-4d41-96b0-69e19240827e)

Abhisheksainii:
Hello, I am working on the Measurement chart
The tooltip part is done.
Had query if the core/charts.dart needs to be moved to measurements folder?

PR #445, opened 2023-11-05
Maintenance PR

12people:
Various minor fixes, partly to get the project to build on macOS with the latest version of Flutter, partly to fix lint warnings.

All changes are annotated in the relevant commits.


12people:
The CI tests all fail because they use an old Dart version.
Could the Dart SDK for those checks be updated?

12people:
I also added new icons for all platforms.
iOS now has sharper (but optimized via optipng) icons, macOS now has squircle icons, as is the norm on latest macOS versions, and Android now has a vector-based icon, which also serves as a monochrome icon (solving the problem that is mentioned in #442 ).

rolandgeider:
Cool, thanks!

Yeah, we need to finish work on `fix/replace-chart-lib`, the old chart lib was discontinued and is preventing us from updating flutter to something not old, but it's just some polishing left (#433)

rolandgeider:
> and Android now has a vector-based icon, which also serves as a monochrome icon (solving the problem that is mentioned in #442 ).

aweome, will check it out

12people:
> Cool, thanks!
> 
> Yeah, we need to finish work on `fix/replace-chart-lib`, the old chart lib was discontinued and is preventing us from updating flutter to something not old, but it's just some polishing left (#433)

Oh, what I did was just replace it with the community-maintained fork: [`community_charts_flutter`](https://pub.dev/packages/community_charts_flutter). It's basically a drop-in replacement.

I do agree, though, that using `fl_charts` is probably best in the long run.

12people:
@rolandgeider Manually.

I made a template for app icons in Figma for myself, which I've used for all my projects and it was easy to just [plug the wger icon into that](https://www.figma.com/file/5APvN3X7Lb1LQkechUPNbg/wger-design?type=design&node-id=1506%3A9793&mode=design&t=5EC7aojTMtPKNQlW-1) too.

Figma lets you specify multiple export sizes as well as suffixes, so once you've got that set up in Figma, getting the iOS and macOS icons is just a simple matter of dragging those exports into the right folder. For Android, I exported an SVG for the foreground and converted that into an Android Drawable via Android Studio.

(I didn't update any of the old Android PNGs, as that would require manually renaming each exported PNG and dragging it into the right folder — while Figma supports specifying export prefixes and suffixes, it does not support specifying different folders for each export, so it's not quite as simple to generate PNG icons for Android. I suppose you could write an automated script for this, but a vector version works well enough for everything v26 and onward, so I didn't see a need to.)

rolandgeider:
so if/when we change the target sdk to 26 we could remove the android png icons, right?

12people:
> so if/when we change the target sdk to 26 we could remove the android png icons, right?

Sure.

That said, the vector drawable in the PR is used automatically here for all platforms 26 and above, so the PNGs only serve as fallback for older platforms.

rolandgeider:
👍🏻 

yeah, that's what I meant. I think at the moment we have something lower than that, like 23 or so

rolandgeider:
Merged thanks :)

PR #473, opened 2023-12-23
Improve flatpak local build

bbkz:

For `flatpak_packager.dart` `localReleaseAssets` in  `flatpak_meta.json`  needs to be empty (kind of). But for `manifest_generator.dart` it needs to be filled correctly otherwise the flatpak manifest will not be correct.

So i handled the data in a seperate json file.

## Checklist

- [-] Tests for the changes have been added (for bug fixes / features)
- [-] Added yourself to AUTHORS.md <- not in application code

### DIFF
--- a/flatpak/scripts/README.md
+++ b/flatpak/scripts/README.md
@@ -9,11 +9,15 @@ If uploading to Flathub for the first time, follow the official contributing gui
 
 To run a local build:
 
-1. Build the Linux release using Flutter.
+1. Build the Linux release using Flutter. `flutter build linux --release`
 2. Add the new release version and date to the start of the "releases" list in the "spec.json" file (and adjust other parameters in the file if needed).
-3. Run "dart flatpak_packager.dart --meta ../flatpak_meta.json -n pubspec.yaml" in this folder.
-4. Run "dart manifest_generator.dart --meta ../flatpak_meta.json -n pubspec.yaml" in this folder.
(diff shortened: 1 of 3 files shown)


rolandgeider:
@12people this is your flatpak script, perhaps something you can use?

12people:
@bbkz @rolandgeider 

Actually, I think the point of the `localReleaseAssets` aspect of the json file was misunderstood.
It's there to define local release assets, which is useful even when building a local Flatpak (out of local files, e.g. for testing).

Whether to use the local assets or to download them from Github instead is defined using the `--github` flag for the script.

The reason why there's one manifest file instead of two is to have one place for specifying all the necessary data instead of two (which could lead to inconsistencies).

12people:
BTW, I've since updated the package files somewhat, for my own project: https://gitlab.com/enjoyingfoss/parlera/-/tree/master/flatpak/scripts?ref_type=heads

Will submit a pull request later when I get to it.

bbkz:
Nice, thank you. Yes, i wanted to build the local flatpak package and added this, for not having to manually add the assets after building the package.

For the local flatpak installation with `flatpak-builder --user --install flatpak-build-dir de.wger.flutter.json`, i had to add this.

Happy to see your latest impelmentation ;-)

rolandgeider:
might also be a good idea to separate the scripts into their own repo (I think we already talked about this once?) and, for the moment, just copy it from there. At least that way we will know if there are any upstream changes

PR #566, opened 2024-04-27
Nutrition cleanup & further enhancements

Dieterbe:

- cleanup nutrition code
- nutrition dashboard widget: use new barchart such that we can see our progress towards the goal
- nutrition dashboard widget: 3 levels of detail (don't show all the meals by default)
- nutrition dashboard widget: show caloric balance
- suggestion: replace chart and horizontal table with vertical table (this one may be contentious / need more discussion), remove FlNutritionalDiaryChartWidget
-  new "progress towards goals" widget for nutrional plans, both for dashboard widget and nutritional details page

## Link to the issue : 

(Add link of the issue you have proposed changes to)

- Link :

## Tests

Please make sure to add tests when implementing new features.

## Checklist


lib/widgets/nutrition/charts.dart
```diff
@@ -19,12 +19,128 @@
...
+    final plan = widget._nutritionalPlan;
+    final today = plan.loggedNutritionalValuesToday;
+
+    return LayoutBuilder(builder: (context, constraints) {
+      var maxVal = 1.0;
+// if any of the bars goes over 100%, find the one that goes over the most
+// that one needs the most horizontal space to show how much it goes over,
+// and therefore reduces the width of "100%" the most, and this width we want
+// to be consistent for all other bars
+      if (plan.goalProtein != null && today.protein / plan.goalProtein! > maxVal) {
```
rolandgeider, line 93: Better use plan.plannedNutritionalValues.protein / fat / etc.

Otherwise we would _only_ use the values from the manual goals, and not the calculated values from the meals

lib/widgets/nutrition/charts.dart
```diff
@@ -19,12 +19,128 @@
...
+    required NutritionalPlan nutritionalPlan,
+  }) : _nutritionalPlan = nutritionalPlan;
+
+  final NutritionalPlan _nutritionalPlan;
+
+  @override
+  State<StatefulWidget> createState() => FlNutritionalPlanGoalWidgetState();
+}
+
+// * fl_chart doesn't support horizontal bar charts yet.
```
rolandgeider, line 39: I think I mentioned it on discord before, but if we have to contort ourselves for such things, we could just use https://pub.dev/packages/graphic for such things, which seems to support this natively (same with time series, which cause a bit of pain with fl)

lib/widgets/dashboard/widgets.dart
```diff
@@ -170,15 +174,19 @@ class _DashboardNutritionWidgetState extends State<DashboardNutritionWidget> {
...
+                    child: switch (_viewMode) {
+                      viewMode.base => const Icon(Icons.info_outline),
+                      viewMode.withMeals => const Icon(Icons.info),
+                      viewMode.withMealsDetails => const Icon(Icons.info),
+                    })
                 : const SizedBox(),
             onTap: () {
               setState(() {
-                _showDetail = !_showDetail;
+                _viewMode = switch (_viewMode) {
```
rolandgeider, line 185: good idea! this definitely looks way cleaner

lib/widgets/nutrition/charts.dart
```diff
@@ -476,192 +584,3 @@ class MealDiaryBarChartWidgetState extends State<MealDiaryBarChartWidget> {
     );
   }
 }
-
-class FlNutritionalDiaryChartWidget extends StatefulWidget {
```
rolandgeider, line 480: Wouldn't this historical view be nice to have?

lib/widgets/nutrition/charts.dart
Dieterbe: I'll have a look at graphic. Though I'm quite pleased with how simple and elegant the "diy gauge" approach is.. I doubt it'll be as straightforward with the graphic package

lib/widgets/nutrition/charts.dart
Dieterbe: I guess it could be useful somewhere, somehow. It's not very practical though. (Various UX issues, see discord). And I don't we should have both the chart *and* the table on the same page

lib/widgets/nutrition/charts.dart
rolandgeider: sure, this was just a suggestion in case we need something else in the future

lib/widgets/nutrition/charts.dart
Dieterbe: good point, but i just noticed an issue with `NutritionalPlan.plannedNutritionalValues`:
* if the planned nutritionalValues come from meals, then a goal of 0 means "this ingredient was not found in any of the mealitems that you've defined, therefore, 0g is your upper limit. if you eat more, you've exceeded your plan.
* if they come from the goals that the user has specified in the nutritional plan settings, than null is translated to 0, and here a goal of 0g means "you've set no restriction on this macro, so you can't hit a limit"

it seems to me this null->0 conversion is a bug. `NutritionalPlan.plannedNutritionalValues` should be able to return null values for goals that are not defined.

PR #584, opened 2024-05-18
Fiber goals

Dieterbe:
Support for fiber goals

if you have it defined as a goal, it'll also show up in a few places automatically, but not everywhere, since it's not as important as the main macros.

this depends on https://github.com/wger-project/wger/pull/1664


lib/widgets/nutrition/forms.dart
```diff
@@ -148,7 +148,6 @@ class MealItemForm extends StatelessWidget {
             IngredientTypeahead(
               _ingredientIdController,
               _ingredientController,
-              showScanner: true,
```
rolandgeider, line 151: why remove the scanner?

rolandgeider: oh yes, you are right

lib/widgets/nutrition/forms.dart
```diff
@@ -447,23 +464,44 @@ class _PlanFormState extends State<PlanForm> {
...
-              if (!value) {
-                widget._plan.goalEnergy = null;
-                widget._plan.goalProtein = null;
-                widget._plan.goalCarbohydrates = null;
-                widget._plan.goalFat = null;
-              }
-            },
+          Row(
+            children: [
+              Text('Macro goals', style: Theme.of(context).textTheme.titleMedium),
```
rolandgeider, line 469: can you make this text translatable?

PR #585, opened 2024-05-19
Add energy and kcal values to ingredient logging/adding forms

Dieterbe:
When:
1) logging an ingredient
2) modifying a meal and adding or an ingredient

the app already shows you a list of recently used ingredients, as well a form with the ingredient, and amount.

what this PR does:
1) on the list of recently used ingredients, show energy and macros of the ingredient on each tile
2) on the form, give a preview of what you're about to log/add to the meal, in terms of energy and macros. updates values in real time as you change the amount.

this should be handy because often i log an entry (save the form), and then check my diary and realize it isn't quite what i wanted. this is especially true when you are dealing with an ingredient on your plate that doesn't have an exact match in the database, and you have to use something "similar": it's nice to be able to see if the values make sense before you log it.

I also added a couple of cleanups and refactors to this PR. notably, the 2 mentioned forms were quite similar, so i merged them into 1 class.

future work: add a warning icon to the list tile in case the ingredient is missing a field that you need to to track against a goal (e.g. fiber)

### DIFF
--- a/lib/models/nutrition/ingredient.dart
+++ b/lib/models/nutrition/ingredient.dart
@@ -18,6 +18,7 @@
 import 'package:json_annotation/json_annotation.dart';
 import 'package:wger/helpers/json.dart';
 import 'package:wger/models/nutrition/image.dart';
+import 'package:wger/models/nutrition/nutritional_values.dart';
 
 part 'ingredient.g.dart';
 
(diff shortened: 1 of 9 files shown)


rolandgeider:
does the preview only show when you select an ingredient from the quicklist?

Dieterbe:
the preview should not show if you don't have any ingredient in the form
once you click on a tile to fill the form, it should appear.
when you scan an ingredient (or use the search function and select a search result) and add the ingredient that way, it probably doesn't show the preview yet.  I can work on that next though

rolandgeider:
> there's a actually some weird behavior around the search/scan function (predating my changes) that i was planning to work on.

🚀

Dieterbe:
see #592 for followup

PR #586, opened 2024-05-21
cleanup for rive workaround

Dieterbe:
flutter always automatically does these changes, any time i build the app.
so i guess they should be committed?


rolandgeider:
Yeah, they probably should be in there

PR #593, opened 2024-05-27
log meal: separate confirmation page with portion choice

Dieterbe:
log meal: separate confirmation page with portion choice

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -38,6 +38,7 @@ import 'package:wger/screens/form_screen.dart';
 import 'package:wger/screens/gallery_screen.dart';
 import 'package:wger/screens/gym_mode.dart';
 import 'package:wger/screens/home_tabs_screen.dart';
+import 'package:wger/screens/log_meal_screen.dart';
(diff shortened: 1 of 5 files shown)


rolandgeider:
Would a slider be better here? Then it's easier to change the portion size (and it's clearer that we're talking about percentages)

Dieterbe:
I was thinking something similar. But then I realized I want to be able to enter, say 150% or 200% for those times I need extra kcals. And any such UI would impose an upper limit

rolandgeider:
I mean, the slider can also go to 200, with the default being 100 (and you can just log twice, if you need once a bigger amount)

rolandgeider:
alright

Dieterbe:
at a quick glance, we don't currently use such a widget anywhere, so i'll find one on pub.dev, unless you have a tip?

rolandgeider:
I think flutter itself has one

rolandgeider:
yep https://api.flutter.dev/flutter/material/Slider-class.html

PR #604, opened 2024-06-10
kcal/macros total for meals + fix alignments

Dieterbe:
Let's use a fixed set of flex factors, corresponding to the kcal and macros columns.

these factors are then used in a bunch of widgets, so that the values align vertically, regardless of their values/contents
ListTile had too many implicit layout behaviors, replaced it with a more straightforward NutritionTile that we use everywhere.
playground example to understand how the layout works: https://dartpad.dev/?id=00c9d9f7d21b03cc3889361415262cc3


rolandgeider:
image

Wouldn't having the ingredient names left-aligned be better?

(also I'd add a little bit of padding at the very bottom, but this is already an improvement)

Dieterbe:
> Wouldn't having the ingredient names left-aligned be better?

i was wondering the same thing.  i guess i'll do it (that's how it is in master too)

PR #606, opened 2024-06-10
Simplify tables code + small fixes and tweaks

Dieterbe:


rolandgeider (approved):

lib/widgets/nutrition/nutritional_diary_detail.dart
```diff
@@ -87,125 +87,53 @@ class NutritionDiaryTable extends StatelessWidget {
...
+          columnHeader(false, loc.difference),
+        ]),
+        macroRow(0, false, loc.energy, (NutritionalValues nv) => nv.energy),
+        macroRow(0, true, loc.protein, (NutritionalValues nv) => nv.protein),
+        macroRow(0, true, loc.carbohydrates, (NutritionalValues nv) => nv.carbohydrates),
+        macroRow(1, true, loc.sugars, (NutritionalValues nv) => nv.carbohydratesSugar),
+        macroRow(0, true, loc.fat, (NutritionalValues nv) => nv.fat),
+        macroRow(1, true, loc.saturatedFat, (NutritionalValues nv) => nv.fatSaturated),
+        macroRow(0, true, loc.fiber, (NutritionalValues nv) => nv.fiber),
+        macroRow(0, true, loc.sodium, (NutritionalValues nv) => nv.sodium),
```
rolandgeider, line 136: FYI, while that doesn't apply in this widget, the flutter team recommends creating new widgets instead of using helper methods https://www.youtube.com/watch?v=IOyq-eTRhvo

Dieterbe: i do think that would make the code more verbose. with the helper method it's really compact and easy to read.

but yeah, we can try out that approach later. let's just merge this now before conflicts accumulate.

rolandgeider: sure. Plus this is only important if we have nested and complex widget trees, which is not the case here

PR #613, opened 2024-06-20
popup ingredient details in typeahead suggestions and recently used

Dieterbe:
unfortunately, we only get the image when getting suggestions. not when using the ingredients API or when browsing recently used ingredients. this would be a nice thing to add to the backend @rolandgeider :) i also think a link to the "upstream source" could be nice. (e.g. to OFF or USDA if applicable)


rolandgeider:
yes, ideally we would query the regular ingredient API, then we would have access to all fields and could decide what to show

rolandgeider:
Just tested this, I like it! However, I am always getting a render flex overflow error here (it seems only the first time?). I would also remove the g-per-body-kg column, that's only useful for the whole plan

Dieterbe:
blocked on https://github.com/wger-project/wger/issues/1713

rolandgeider:
> blocked on [wger-project/wger#1713](https://github.com/wger-project/wger/issues/1713)

have updated the test and the production server

lib/widgets/nutrition/ingredient_dialogs.dart
```diff
@@ -0,0 +1,199 @@
...
+
+    return AlertDialog(
+      title: (snapshot.hasData) ? Text(ingredient!.name) : null,
+      content: SingleChildScrollView(
+        child: Padding(
+          padding: const EdgeInsets.all(8.0),
+          child: Column(
+            mainAxisSize: MainAxisSize.min,
+            children: [
+              if (snapshot.hasError)
```
rolandgeider, line 47: I feel we could rewrite this list so that we only have one snapshot.hasData (and .hasError) check and then show all the ingredient and not need to check for hasData on every line, but it works, so 🤷🏻‍♂️

lib/widgets/nutrition/forms.dart
```diff
@@ -353,7 +353,7 @@ class IngredientFormState extends State<IngredientForm> {
                 child: Column(
                   children: [
                     Text(
-                      'Macros preview',
+                      'Macros preview', // TODO fix l10n
```
rolandgeider, line 356: l10n :)

lib/widgets/nutrition/ingredient_dialogs.dart
```diff
@@ -0,0 +1,199 @@
...
+                ConstrainedBox(
+                  constraints: const BoxConstraints(minWidth: 400),
+                  child: MacronutrientsTable(
+                    nutritionalGoals: goals!,
+                    plannedValuesPercentage: goals.energyPercentage(),
+                    showGperKg: false,
+                  ),
+                ),
+              if (snapshot.hasData && ingredient!.licenseObjectURl == null)
+                Text('Source: ${source!}'),
```
rolandgeider, line 65: l10n

lib/widgets/nutrition/ingredient_dialogs.dart
```diff
@@ -0,0 +1,199 @@
...
+                ConstrainedBox(
+                  constraints: const BoxConstraints(minWidth: 400),
+                  child: MacronutrientsTable(
+                    nutritionalGoals: goals,
+                    plannedValuesPercentage: goals.energyPercentage(),
+                    showGperKg: false,
+                  ),
+                ),
+              if (ingredient != null && ingredient.licenseObjectURl == null)
+                Text('Source: ${source!}'),
```
rolandgeider, line 165: l10n

PR #622, opened 2024-07-11
Use flutter 3.22 - upgrade intl, rerun codegen, upgrade some dependencies

Dieterbe:


rolandgeider:
Can you also bump the version in .github/workflows?

Issue #628, opened 2024-08-14
Set weight interface seems clumsy


ysalmon:
## Steps to Reproduce

1. open wger app for android
2. in the Weight section, click +

In addition :
3. click the ``-`` or ``+`` around the displayed weight.

**Expected results:** 
1.2. The selected date for adding a weight measurement should be today.

3. Increment or decrement in weight should be practical, ie. either 0.1 kg or 0.05 kg.

**Actual results:** 
1.2. The selected date is the day before (maybe because that was the day for which a measurement was last stored).

3. The weight increment/decrement is 0.25 kg, which is impractical when your scale gives measurements in steps of 0.1 kg.

rao-suraj:
I can work on this if you want it to get changed.

Dieterbe:
this was bothering me too. the fix has been merged, and should appear in the playstore/appstore soon

PR #632, opened 2024-09-05
Weight entry QoL improvements

sizzlesloth:

- Use current DateTime when copying WeightEntry instead of using previous weight.
- Increment buttons increment weight by 0.1 and round to one decimal place to prevent floating point accuracy (rounding) issues.

## Link to the issue : 

- Link : #628 

## Checklist

- [X] Tests for the changes have been added (for bug fixes / features)
- [X] Added yourself to AUTHORS.md

### DIFF
--- a/lib/models/body_weight/weight_entry.dart
+++ b/lib/models/body_weight/weight_entry.dart
@@ -43,7 +43,7 @@ class WeightEntry {
   WeightEntry copyWith({int? id, int? weight, DateTime? date}) => WeightEntry(
         id: id,
         weight: weight ?? this.weight,
-        date: date ?? this.date,
+        date: date ?? DateTime.now(),
       );
 
   // Boilerplate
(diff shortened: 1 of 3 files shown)


lib/models/body_weight/weight_entry.dart
```diff
@@ -43,7 +43,7 @@ class WeightEntry {
   WeightEntry copyWith({int? id, int? weight, DateTime? date}) => WeightEntry(
         id: id,
         weight: weight ?? this.weight,
-        date: date ?? this.date,
+        date: date ?? DateTime.now(),
```
Dieterbe, line 46: when you call copyWith(), you expect that it copies the model, and overrides those fields that are given.
therefore this change doesn't make sense. Instead, the callers should just provide whichever date they want to use (DateTime.now())

Dieterbe (changes requested):

Dieterbe:
also `flutter test` is broken.

your PR is a good start, it just needs some tweaks. i will finish it.

Dieterbe:
@sizzlesloth thanks! i merged your changes via the PR above.

sizzlesloth:
Thank you very much! :)

PR #708, opened 2025-01-03
Adds ability to use API Key to "sign in"

eyJhb:
- Adds a API Key to the sign in screen

## Link to the issue : 
- Link: https://github.com/wger-project/wger/issues/1847

## Tests
Unsure how to do this.

## Checklist

image


Dieterbe:
Why do we want this?

eyJhb:
> Why do we want this?

Look at the linked issue here https://github.com/wger-project/wger/issues/1847 , without this they wouldn't be able to use the app at all. Since the idea is to set up a 3rd party auth service in front, e.g. Authelia, which ensures that the user is authenticated and passes it on to Wger. 

This in turn means, that the user does not have a username/password combination that works for Wger (only a username), and they would be unable to sign in using the app. Instead they can provide the API token from the settings page, and use that instead.

Dieterbe:
interesting. thanks. so i suppose this would also require a way to register users with wger and instantiate api keys for them somehow.

eyJhb:
That's taken care of in this PR https://github.com/wger-project/wger/pull/1859 , when enabled, Wger should assume that any request with e.g. `Remote-User: eyjhb` in the header, is authenticated and will then create the user.

So in the case for the app, that wouldn't be required/make any sense to implement the part for registering, as that would be outside the scope of the app, in the case of having that specific setup. The only way to implement it, would be to open a webview of the instance URL, let the user authenticate, and then point them to the API key page. But I think that is out of scope. I think it's expected, that with a setup that uses Auth Proxy, that registering using the app is not possible :)

eyJhb:
But it would maybe be nice, to hide it under an "advanced options", together with custom server URL as well.

rolandgeider:
> But it would maybe be nice, to hide it under an "advanced options", together with custom server URL as well.

yes, definitely. I can help you with that (and with the tests)

eyJhb:
I might have time to look at it, but feel free to take over the PR. I just made it to get the ball rolling, and show the basic idea that I have :)

rolandgeider:
I'm thinking if it makes sense to just also enter the api key in the password field. If the username + password combo doesn't work we try sending content as the token in the header

rolandgeider:
Used your solution with the extra field, but made the token toggle only visible when showing the custom server, since this is the only way when using a token directly is useful:

Simulator Screenshot - iPhone 16 Plus - 2025-04-30 at 22 45 48
Simulator Screenshot - iPhone 16 Plus - 2025-04-30 at 22 45 51
Simulator Screenshot - iPhone 16 Plus - 2025-04-30 at 22 45 53

PR #781, opened 2025-04-04
Add swipe-to-delete and improve UI design (for Routines, Nutrition Plans Screens)

Hamed233:
# Proposed Changes

- Added swipe-to-delete functionality using `Dismissible` widget in both Routines and Nutrition Plans.
- Improved UI/UX of routine and plan cards:
  - Rounded corners, shadows, consistent spacing
  - Better typography and color contrast
- Retained delete icon for manual deletion with confirmation dialogs
- Snackbar feedback after successful deletion


Dieterbe:
> Added swipe-to-delete functionality using Dismissible widget in both Routines and Nutrition Plans.

in my android email app (bluemail), i sometimes accidentally delete stuff when i try to swipe up/down but it's not perfectly vertical. so i hope we don't introduce such issues here

Issue #793, opened 2025-04-17
Allow editing an existing session


rolandgeider:
Users should be able to edit the sessions in the workout log view:

![Image](https://github.com/user-attachments/assets/19e4db80-1176-4ed3-bdb4-31a3ea0dbe61)

Rohitdhall987:
Hi! I’m a beginner and would love to contribute to this. Can I take this issue?

rolandgeider:
hi @Rohitdhall987, sure go ahead!

If you have any problems or questions, just ping me :)

rolandgeider:
Just forgot, in the `GymMode` there's a session form to create new ones, we could probably re use some of that

Rohitdhall987:
Hi @rolandgeider, I have a few questions before I begin.

Based on the screenshot you provided at the top of the issue, it looks like on the dashboard there's a calendar at the bottom that shows the workout sessions for each date, including their start/end time and rating.

While testing the app, I noticed that the session timing and workout rating are already editable — but only for the current date. 

So I just wanted to clarify:
👉 Do you want users to be able to edit **past sessions** as well?
Or did I perhaps misunderstand the intended enhancement?

Thanks in advance!

rolandgeider:
oh sorry, _our_ calendar is on the log page of a routine 😅

<img width="329" alt="Image" src="https://github.com/user-attachments/assets/d136f8a3-033a-4f54-a197-dc9f24044d73" />

But yes, the idea is to be able to edit any past sessions the user might have

Also forgot to tell you, you can generate a session by e.g. starting the gym mode (the "play" button on the routine days) and saving a log entry. If there's no session attached to that day, a new one is created automatically by the backend, you don't need to reach the last page and save it there.

rolandgeider:
merged @Rohitdhall987 , thanks

PR #798, opened 2025-04-25
feat: allow editing old sessions via popup dialog

Rohitdhall987:
# Proposed Changes

- Added a popup dialog to edit workout session data directly.
- Used a simplified approach to avoid complex prop drilling and tightly-coupled widget structure.

## Related Issue(s)

- See also: #793

## Description

The existing code structure was quite nested and involved a lot of prop drilling.  
Many parts of the UI depended on each other, and attempting to refactor everything to insert editing logic would risk breaking the flow or functionality.

Instead of trying to untangle the entire structure, I went for a simpler and more isolated approach by adding an editable popup dialog.  
This keeps the current structure intact while still allowing the session to be updated efficiently.

I reused your existing logic and helper functions wherever possible to keep the behavior consistent.  
Functionally, the popup works identically to the original workflow but with less complexity.

### DIFF
--- a/lib/widgets/routines/log.dart
+++ b/lib/widgets/routines/log.dart
@@ -18,13 +18,16 @@
 
 import 'package:flutter/material.dart';
 import 'package:intl/intl.dart';
+import 'package:provider/provider.dart';
 import 'package:wger/helpers/colors.dart';
+import 'package:wger/helpers/json.dart';
 import 'package:wger/helpers/misc.dart';
 import 'package:wger/helpers/ui.dart';
 import 'package:wger/l10n/generated/app_localizations.dart';
 import 'package:wger/models/workouts/log.dart';
 import 'package:wger/models/workouts/routine.dart';
 import 'package:wger/models/workouts/session.dart';
+import 'package:wger/providers/routines.dart';
 import 'package:wger/widgets/measurements/charts.dart';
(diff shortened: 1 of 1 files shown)


rolandgeider:
> The existing code structure was quite nested and involved a lot of prop drilling

that is sadly very true (even if it's gotten better)

lib/widgets/routines/log.dart
```diff
@@ -197,3 +207,114 @@ class DayLogWidget extends StatelessWidget {
...
 }
+void showEditSessionDialog(BuildContext context, WorkoutSession session, RoutinesProvider routinesProvider) {
+  final _formKey = GlobalKey<FormState>();
+  final notesController = TextEditingController(text: session.notes);
+  final timeStartController = TextEditingController(text: timeToString(session.timeStart ?? TimeOfDay.now())!);
+  final timeEndController = TextEditingController(text: timeToString(session.timeEnd ?? TimeOfDay.now())!);
+  List<bool> selectedImpression = [false, false, false];
+  selectedImpression[session.impression - 1] = true;
+
+  showDialog(
```
rolandgeider, line 218: I had thought we could refactor `SessionPage` a bit, extract from there the form logic to its own form widget and simply reuse that here as well? Then we don't have two places that edit sessions

PR #814, opened 2025-05-08
Ingredient ordering

### LINKED ISSUE #805: Ordering of previous ingredients
### Use case

The list item of previously eaten ingredients should be better ordered.

It should reverse the current date order, to newest at the top - In the least.

Better still it should arrange by frequency. Putting regular choices at the top 

### Proposal

_No response_

### Additional Context

_No response_

Shininess8816:
# Proposed Changes

Add frequency of logged ingredients + Sort by descending

## Related Issue(s)

Closes #805

### DIFF
--- a/lib/models/nutrition/nutritional_plan.dart
+++ b/lib/models/nutrition/nutritional_plan.dart
@@ -16,6 +16,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import 'dart:collection';
 import 'package:collection/collection.dart';
 import 'package:flutter/widgets.dart';
 import 'package:json_annotation/json_annotation.dart';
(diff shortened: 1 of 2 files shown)


lib/models/nutrition/nutritional_plan.dart
```diff
@@ -226,15 +227,25 @@ class NutritionalPlan {
...
-      );
-      if (found == null) {
-        out.add(log);
-      }
+      final int howMany = diaryEntries
+          .where((diaryEntries) =>
+              diaryEntries.ingredientId == log.ingredientId &&
+              diaryEntries.amount == log.amount &&
+              diaryEntries.datetime.isAfter(DateTime.now().add(recentWindow)))
+          .length;
```
Dieterbe, line 239: hmm we do (diaryEntries.length)^2 iterations here. could be a performance issue. i have a lot of diary entries.
perhaps we could iterate once and update a map of ingredientId->count records.

not sure what's the best way to compare the performance though, does dart come with benchmark tools?

Issue #784, opened 2025-04-07
Android App - Timer doesn't work in the background


burny0202:
### Priority/Impact

None

### Description

When the app is minimized or the screen locked - The timer in between sets does not continue

### Mobile app version

1.8

shraavv:
Hi, I noticed that the timer between sets works fine on my phone, even when I minimize the app or lock the screen (though there’s a slight lag). While looking into this, I found that on some devices this issue can occur due to how the OS handles background processes and timers. In timer.dart, we're currently using an int variable to track elapsed seconds with Timer.periodic, which isn't the most reliable approach for timekeeping. A better and recommended solution I found would be to switch to using a DateTime start time and calculate the elapsed duration using DateTime.now().difference(startTime). This way, the timer would accurately update the time. Shall I just update this?

Shininess8816:
Personally I'd like it to wake screen. I don't want to sit staring at this page to wait for 'go'

shraavv:
> I have a solution that involves using local storage options such as Isar, Drift, Hive, or SharedPreferences, depending on the complexity of the data being stored. I highly recommend using Drift due to its efficiency and support for ACID transactions. Specifically, we can create a local storage entry when the timer starts and calculate the elapsed time by comparing the current time with the stored time.
> 
> ![Image](https://github.com/user-attachments/assets/dec70814-341e-486a-829d-4e0f1b399c26)

oh yes, this seems like a good alternative too, will try it out
thanks!

shraavv:
> Personally I'd like it to wake screen. I don't want to sit staring at this page to wait for 'go'

could you please elaborate? i did not get your point

rolandgeider:
> A better and recommended solution I found would be to switch to using a DateTime start time and calculate the elapsed duration using DateTime.now().difference(startTime).

@shraavv yes, this sounds like a good solution. Note that there is a regular timer and a count down timer, depending on whether the exercise has a rest time configured. I also would just use the gym provider and not save anything anywhere, if the user completely closes the app, this progress is lost, but that's ok.

PR #838, opened 2025-06-09
timer code is now updated and uses a better approach

shraavv:
# Proposed Changes

The timer issue faced could be due to how the OS handles background processes and timers. The issue could be addressed with a better solution which uses  DateTime start time and calculate the elapsed duration using DateTime.now().difference(startTime). This way, the timer would accurately update the time and the OS issue might be solved too.

## Related Issue(s)

Closes issue #784


rolandgeider:
thanks for the PR, and sorry for the wait!

PR #854, opened 2025-06-27
fix rive build on linux.

Dieterbe:
see https://github.com/rive-app/rive-flutter/issues/482


SalehTZ:
Is this a fix for #390 and #482 ?

PR #855, opened 2025-06-27
Nutrition plan stats

Dieterbe:
support for plan start & end dates.
show stats on NP list page

image

Needs wger-project/wger#1997


lib/widgets/nutrition/forms.dart
```diff
@@ -518,6 +520,13 @@ class _PlanFormState extends State<PlanForm> {
 
     _onlyLogging = widget._plan.onlyLogging;
     _descriptionController.text = widget._plan.description;
+    _startDateController.text =
+        '${widget._plan.startDate.year}-${widget._plan.startDate.month.toString().padLeft(2, '0')}-${widget._plan.startDate.day.toString().padLeft(2, '0')}';
```
rolandgeider, line 524: you can just use `dateToYYYYMMDD` for this

PR #928, opened 2025-09-27
Multi nutrition plan fixes

Dieterbe:
# Proposed Changes

-
-

## Related Issue(s)

If applicable, please link to any related issues (`Closes #123`,
`Closes wger-project/other-repo#123`, `See also #123`, etc.)

  (run `dart format .`)

image

### DIFF
--- a/test/nutrition/nutritional_plans_screen_test.dart
+++ b/test/nutrition/nutritional_plans_screen_test.dart
@@ -152,15 +152,15 @@ void main() {
     await tester.pumpWidget(createHomeScreen());
 
 // note .. "(open ended)" at the time, depending on localisation strings
-    expect(find.textContaining('from 1/1/2021 ('), findsOneWidget);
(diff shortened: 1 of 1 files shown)


lib/helpers/date.dart
```diff
@@ -30,6 +30,51 @@ List<DateTime> daysInRange(DateTime first, DateTime last) {
   );
 }
 
+/// Formats a Duration into a human-readable string with years, months, and days
+String humanDuration(DateTime startDate, DateTime endDate) {
```
rolandgeider, line 34: I would just use an external package for this (one of the very first results was https://pub.dev/packages/humanize_duration), then we would also have i18n solved

Dieterbe: these types of libraries don't do it the way i think makes sense.
e.g. day A you reported a weight at 9:20 AM, one day later you weigh in at 9:00 AM, i think that should qualify as "a day"

PR #978, opened 2025-11-04
Save ingredient to cache on selection

### LINKED ISSUE #932: Save ingredient to cache on selection
### Use case

Currently when the user searches for an ingredient, e.g. when logging a single ingredient to a meal, we need to re-fectch it so it gets saved to the cache.

### Proposal

When an ingredient is selected on the autocomplete (see uses of NutritionPlansProvider.searchIngredient), we should save it to the cache. We would refactor `fetchIngredient` and extract the logic to save the ingredient to the cache to a new function so that we can use it in the autocomplete.

### Additional Context

_No response_

lenka369:
# Proposed Changes
- Refactored fetchIngredient to extract the ingredient caching logic into a separate cacheIngredient function.
- Added a cacheIngredient call to the autocomplete’s onSelected callback so that ingredients are cached immediately after selection.
- Added unit and widget tests to verify that cacheIngredient is called when selecting an ingredient from the autocomplete dropdown and that it saves ingredients to both the in-memory list and the database cache.

## Related Issue(s)

If applicable, please link to any related issues (`Closes #932 `)

  (run `dart format .`)

### DIFF
--- a/lib/providers/nutrition.dart
+++ b/lib/providers/nutrition.dart
@@ -306,19 +306,31 @@ class NutritionPlansProvider with ChangeNotifier {
   Future<void> cacheIngredient(Ingredient ingredient, {IngredientDatabase? database}) async {
     database ??= this.database;
 
-    ingredients.add(ingredient);
+    if (!ingredients.any((e) => e.id == ingredient.id)) {
+      ingredients.add(ingredient);
+    }
 
-    final data = ingredient.toJson();
-    await database
-        .into(database.ingredients)
-        .insert(
-          IngredientsCompanion.insert(
-            id: ingredient.id,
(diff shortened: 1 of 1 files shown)


lib/providers/nutrition.dart
```diff
@@ -302,6 +302,25 @@ class NutritionPlansProvider with ChangeNotifier {
     await database.deleteEverything();
   }
 
+  /// Saves an ingredient to the cache
+  Future<void> cacheIngredient(Ingredient ingredient, {IngredientDatabase? database}) async {
+    database ??= this.database;
```
rolandgeider: since we call this directly after receiving results from the search, it could happen that the ingredient is already in the db. We should add a check here.

Issue #1071, opened 2026-01-11
[Bug] App crashes with "WgerHttpException (ErrorType.html)" (404) after creating a plan on self-hosted instance


Ririwin:
### Priority/Impact

High (critical issue, blocks workflow)

### Description

I am running a self-hosted instance of wger (Docker on Synology NAS). The server is working perfectly via the web interface (PWA) and direct API calls return 200 OK. However, in the Flutter mobile app, immediately after creating a Nutrition Plan or Workout Plan, the app crashes with a WgerHttpException displaying an HTML 404 page from my reverse proxy/NAS instead of handling a JSON error or a missing resource gracefully.

**Steps to Reproduce**

1.     Log in to the mobile app with a self-hosted instance (https://my-domain.com).
2.     Go to 'Nutrition' or 'Workout'.
3.     Create a new plan.
4.     Close and restart the app (or sometimes immediately after syncing)
5.     The app displays a red screen with the error below.

**Expected behavior**
The app should handle missing resources gracefully or sync correctly without crashing, as the API endpoints are reachable via browser.

**Screenshots & Logs**
Error displayed in the app:
`Error Message: WgerHttpException (ErrorType.html): {html_error: <!DOCTYPE html>
<html>
<head>... (Synology 404 page content) ...
<div class="circle_text">404</div>
...
Stack Trace:
#0      WgerBaseProvider.fetch (package:wger/providers/base_provider.dart:69)
#1      WgerBaseProvider.fetchPaginated (package:wger/providers/base_provider.dart:83)
...
#7      _HomeTabsScreenState._loadEntries (package:wger/screens/home_tabs_screen.dart:102)
`
**Server verification (Browser):**
GET https://my-domain.com/api/v2/version/ -> Returns 200 OK ("2.4.0a2").
GET https://my-domain.com/api/v2/userprofile/ -> Returns valid JSON.
Server logs (Nginx/Gunicorn) show successful 200 responses for /api/v2/login/ and /api/v2/userprofile/ just before the error.

**Environment:**

-     App Version: 1.9.4
-     Server Version: 2.4.0a2 (Docker image wger/server:latest)
-     Platform: Android 
-     Hosting: Synology NAS (Container Manager) behind Reverse Proxy.

**Additional Context**
It seems the app is requesting a resource that does not exist or is malformed, causing my Reverse Proxy (Synology) to return its default 404 HTML page. The app tries to parse this HTML as JSON and throws the exception.
[...]

PR #1073, opened 2026-01-12
Fix: Add missing trailing slashes for API calls (#1071)

Ririwin:
# Proposed Changes
Add missing trailing slashes to API endpoint paths in lib/providers/exercises.dart to prevent 301 redirects on Synology NAS servers.

The affected endpoints are:
- exerciseSearchPath
- categoriesUrlPath  
- musclesUrlPath
- equipmentUrlPath
- languageUrlPath
I will continue to test the software and see if there are any other bugs of this kind.

### DIFF
--- a/lib/providers/exercises.dart
+++ b/lib/providers/exercises.dart
@@ -47,14 +47,14 @@ class ExercisesProvider with ChangeNotifier {
   static const EXERCISE_CACHE_DAYS = 7;
   static const CACHE_VERSION = 4;
 
-  static const exerciseUrlPath = 'exercise';
-  static const exerciseInfoUrlPath = 'exerciseinfo';
-  static const exerciseSearchPath = 'exercise/search';
+  static const exerciseUrlPath = 'exercise/';
+  static const exerciseInfoUrlPath = 'exerciseinfo/';
+  static const exerciseSearchPath = 'exercise/search/';
+  static const categoriesUrlPath = 'exercisecategory/';
+  static const musclesUrlPath = 'muscle/';
+  static const equipmentUrlPath = 'equipment/';
+  static const languageUrlPath = 'language/';
 
-  static const categoriesUrlPath = 'exercisecategory';
-  static const musclesUrlPath = 'muscle';
-  static const equipmentUrlPath = 'equipment';
-  static const languageUrlPath = 'language';
 
   List<Exercise> exercises = [];
 


rolandgeider:
I understand the error you are getting, I just don't understand why it happens! 😄 

On my end the app seems to be adding trailing slashes everywhere it should (this is from a clean startup, so what you were running into):

Bildschirmfoto 2026-01-13 um 15 58 37

Bildschirmfoto 2026-01-13 um 16 06 12

You said that only happens sometimes?

rolandgeider:
thanks for the logs. ~You mean `/api/v2/exerciseinfo` right? `/api/v2/exercise` isn't used anywhere~ (nevermind, it is used to retrieve the IDs). And the exercise endpoint does have a correct slash at the end. That's what I don't understand, from what I can see, every url that hits the api is passed through that `makeUri` helper.

Even the pagination links (which we use without changing anything), do have the slash

Bildschirmfoto 2026-01-14 um 15 02 14

(also, while this won't help us in this case, I've fixed the html error detection we already have so that the error dialog doesn't show the raw html, but renders the html)

rolandgeider:
no, sorry, I meant that I *can't* confirm it! Every request that I'm seeing here does have a slash

Ririwin:
Hi Roland,

I dug deeper into my server logs and found the smoking gun. You were right: this is an environment issue where my Synology Reverse Proxy behaves erratically.

Here is what happens during a single app launch:

1. Most requests reach the Wger container correctly (Log from Wger Docker container):

```
172.18.0.1 - - [15/Jan/2026:08:01:42 +0000] "POST /api/v2/login/ HTTP/1.1" 200 52 "-" "wger App/1.9.4 (de.wger.flutter; build: 140) - https://github.com/wger-project" "192.168.1.1"
172.18.0.1 - - [15/Jan/2026:08:01:42 +0000] "GET /api/v2/version/ HTTP/1.1" 200 9 "-" "Dart/3.10 (dart:io)" "192.168.1.1"
172.18.0.1 - - [15/Jan/2026:08:01:42 +0000] "GET /api/v2/min-app-version/ HTTP/1.1" 200 7 "-" "Dart/3.10 (dart:io)" "192.168.1.1"
172.18.0.1 - - [15/Jan/2026:08:01:42 +0000] "GET /api/v2/version/ HTTP/1.1" 200 9 "-" "Dart/3.10 (dart:io)" "192.168.1.1"
172.18.0.1 - - [15/Jan/2026:08:01:42 +0000] "GET /api/v2/language/ HTTP/1.1" 200 1692 "-" "wger App/1.9.4 (de.wger.flutter; build: 140) - https://github.com/wger-project" "192.168.1.1"
172.18.0.1 - - [15/Jan/2026:08:01:42 +0000] "GET /api/v2/userprofile/ HTTP/1.1" 200 689 "-" "wger App/1.9.4 (de.wger.flutter; build: 140) - https://github.com/wger-project" "192.168.1.1"
```

2. BUT some requests "miss" the proxy rule and hit the Synology default web server instead (Log from Synology System Nginx):

`2026/01/15 08:01:42 [error] 12464#12464: *89662 "/var/services/web/api/v2/language/index.html" is not found (2: No such file or directory), client: 192.168.1.1, server: , request: "GET /api/v2/language/?limit=20&offset=20 HTTP/1.1", host: "wger.domain.com"`

The Synology Nginx tries to serve /api/v2/language/ as a static file from its disk instead of proxying it to Docker, resulting in the 404 HTML page.

I'll close this PR as the missing slash wasn't the root cause (the failing request actually had the slash!). Thanks for pointing me in the right direction to debug my server. I'll look forward to the Dio robustness improvements!

PR #1096, opened 2026-01-28
BUGFIX: unfocus while typing in gym_mode form

### LINKED ISSUE #1090: Weird keyboard behavior when saving training logs sets
### Priority/Impact

High (affects core functionality)

### Description

I don't think this should be the correct behavior when saving reps. Anytime I try to type reps, as soon as one character is inserted the keyboard is hidden. A little annoying.  The behavior should be to wait until the user has typed the reps then the user can hide the keyboard with the back button or scroll down to save

https://github.com/user-attachments/assets/0b07028a-fcc4-4449-8e05-eb330632f8c0

### Server version

_No response_

### Mobile app version

_No response_

yontank:
The Problem:
- User clicks on a the input with a key `Rep-Input-12KG`
- User enters a value `5`
- Since its a stateful widget, a new input is created (i think) with a key `Rep-Input-125KG`
- Previous build doesn't know where `Rep-Input12KG` is, so it drops its focus

I'm still new to flutter, so I'll probably try to write tests next week or so 

# Proposed Changes
- Change Flutter key values to static values

## Related Issue(s)
 Closes #1090 

  (run `dart format .`)

### DIFF
--- a/lib/widgets/routines/gym_mode/log_page.dart
+++ b/lib/widgets/routines/gym_mode/log_page.dart
@@ -251,7 +251,7 @@ class LogsRepsWidget extends ConsumerWidget {
           child: TextFormField(
             decoration: InputDecoration(labelText: i18n.repetitions),
             enabled: true,
-            key: ValueKey('reps-field-$repText'),
+            key: const ValueKey('reps-field'),
             initialValue: repText,
             keyboardType: textInputTypeDecimal,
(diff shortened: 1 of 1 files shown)


rolandgeider:
For such things you don't need to write test, I can imagine that checking that the focus doesn't get lost will be a nightmare (but if you want to try, please do! 😄)

PR #1106, opened 2026-02-03
add sort datetime desc when showing ingredient suggestion

### LINKED ISSUE #1105: Please add add sort for ingredient suggestion
### Use case

Hi, its would be better if there is desc datetime sort for showing ingredient suggestion on log ingredient page
Image

### Proposal

Most likely the suggestion is showing the most recent used or most used item, but since we doesnt have ingredient used count so it okay to use date desc, its reduce scrolling time when inputting same item again, since the suggestion already sorted with the most recent on top

### Additional Context

_No response_

cryptbrn:
# Proposed Changes

- Add date desc sort for log item in nutrition plan model

## Related Issue(s)
- Closes #1105

### DIFF
--- a/lib/models/nutrition/nutritional_plan.dart
+++ b/lib/models/nutrition/nutritional_plan.dart
@@ -261,6 +261,7 @@ class NutritionalPlan {
         out.add(log);
       }
     }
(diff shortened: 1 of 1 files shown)


rolandgeider:
yeah, this makes sense

PR #1136, opened 2026-03-03
Add sanity check for server configuration

### LINKED ISSUE #1116: Add sanity check for server configuration
For self-hosted backends, it can happen that the reverse proxy is not configured correctly. When this happens the initial requests to e.g. `https://wger.my-domain.com:1234/api/v2/endpoint` work well, but the pagination links `next` and `previous` are constructed by the backend from these headers so in case of a misconfiguration these could look like `http://localhost/api/v2/endpoint?limit=...` which will not work (or they might work in the home lab but not outside, etc. etc.).

Idea: add a sanity check after login (perhaps similar to the min server version check) that tries to hit an endpoint like the exercise one and checks that the domain returned in the next link is the same as the one used to login and if not show some information to the user.

alexisbirling-boop:
# Proposed Changes

- Add automatic server configuration check after user login
- Detect reverse proxy misconfigurations by comparing pagination URLs with base URL
- Display warning dialog to users when misconfiguration is detected (e.g., pagination links pointing to `localhost` instead of actual domain)
- Implement graceful error handling to avoid blocking login on check failures
- Add internationalization support for user-facing dialog messages
- Add unit tests covering various misconfiguration scenarios

## Related Issue(s)

Closes #[1116](https://github.com/wger-project/flutter/issues/1116)

## Screenshot of the sanity check widget (while testing)
issue3


lib/widgets/core/server_config_warning_dialog.dart
```diff
@@ -0,0 +1,48 @@
...
+/// Shows a warning dialog when server misconfiguration is detected
+void showServerConfigWarning(BuildContext context) {
+  final i18n = AppLocalizations.of(context);
+
+  showDialog(
+    context: context,
+    builder: (context) => AlertDialog(
+      title: Row(
+        children: [
+          const Icon(Icons.warning, color: Colors.orange, size: 28),
```
rolandgeider: better use something like `Theme.of(context).colorScheme.error` instead of hard coding the color, then this updates automatically if we change something there

rolandgeider:
It seems a lot of commits got into this PR, perhaps because of a merge or rebase? (I always hate it when that happens!) In any case, I've cherry-picked the relevant commits to #1188 so nothing got lost

Closing here

PR #1175, opened 2026-04-13
fix: use unit IDs directly in Log.fromSetConfigData to prevent null on save

### LINKED ISSUE #1149: Cannot save exercise when added outside of the routine
### Priority/Impact

### Description

I have a issue with adding exercises when logging a workout. When I try to add a extra exercise in my routine the app shows a error and does not log anything.

Step by step:
1. I have a routine created
2. Start the routine/log for the day
3. Add a exercise
4. Fill the reps and weight
5. Error
Image

After some digging I found that the issue is with missing units for reps and weight, this is a log from the server
```
django.core.exceptions.ValidationError: ['Repetitions unit must be present if repetitions have a value.']
```

It took me some time to find this, a default value should be set. That should resolve this issue and also time for others experiencing this.

### Server version

2.5.0a1+git6715b84

### Mobile app version

1.10.2

efrahim20319:
When adding an extra exercise during a workout, addExerciseAfterPage() creates a SetConfigData with no unit objects loaded. Log.fromSetConfigData was using the weightUnit/repetitionUnit setters which overwrite the default IDs with null when the objects are null, causing the server to reject the request with a ValidationError ("Repetitions unit must be present").

Fixes #1149

# Proposed Changes

-
-

### DIFF
--- a/lib/models/workouts/log.dart
+++ b/lib/models/workouts/log.dart
@@ -120,11 +120,13 @@ class Log {
 
     weight = setConfig.weight;
     weightTarget = setConfig.weight;
-    weightUnit = setConfig.weightUnit;
+    weightUnitId = setConfig.weightUnitId ?? WEIGHT_UNIT_KG;
+    weightUnitObj = setConfig.weightUnit;
 
     repetitions = setConfig.repetitions;
     repetitionsTarget = setConfig.repetitions;
-    repetitionUnit = setConfig.repetitionsUnit;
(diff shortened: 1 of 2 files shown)


rolandgeider:
looks good, thanks!

Those weightUnit and repetitionUnit setters are a bit of a pain / fragile, we should probably refactor how those work when we merge https://github.com/wger-project/wger/pull/2275

PR #1181, opened 2026-04-21
WIP: Default weight unit

### LINKED ISSUE #1180: Use plate selection to imply default weight unit
### Use case

Currently in gym mode, when no weight is specified for a set in a routine, kg is selected by default, irrespective of the plates settings configured in the app. This means that for every log for a blank weight entry, someone using lbs needs to first hit the toggle to enter weight unit then select lbs from the drop down. 

We should be able to infer what weight unit to use based on the plates setting, but I could see an argument for two distinct settings. 

### Proposal

When working out, I don't typically pre-plan the actual weight, just the reps, adjusting weight per history and conditions. If trying to minimize downtown between sets, adjusting the weight unit for every set creates unnecessary friction and cognitive load. Implying or allowing the default to be configured app wide will improve the overall experience and prevent logging the wrong weight unit accidentally 

### Additional Context

_No response_

jmccaull:
Initial PR to solicit feedback

# Proposed Changes

- refactor main app to provide PlateCalculatorNotifier as a dependency for RoutineProvider
- remove default weight unit from set config data
- adjust RoutineProvider.defaultWeightUnit to use PlateCalculatorNotifier's metric setting
- adjust RoutineProvider logic to utilize defaultWeightUnit when none provided

## Related Issue(s)

Closes #1180 


jmccaull:
Its been a while since I've done any dart and Riverpod is all new to me so I'm not sure if this is the right way to handle the provider setup. I did not think it made sense to refactor all the providers needed for RoutineProvider to be fully Riverpod in a single pull request.

I'm also looking for feedback on the set_config_data changes. I personally don't think DTO like objects should be selecting a default value for a field that is indicated as null-able, but it seems like an established pattern here. If this change goes forward, does it make sense to also refactor the recent log changes here?

rolandgeider:
Hi! Sorry for the late reply.

There's another PR that I'm working on that is migrating the providers to riverpod, so that problem will disappear in the future.

And as for the PR, I'm not sure if making the plate calculator the source of truth is the correct way. There is already a flag in the profile that handles this (but as you've probably noticed, hasn't the biggest impact in the app) and I would tend to use that (also for the initial value for the plate calculator itself).

jmccaull:
@rolandgeider It looks like the only place the user profile is used is within the home tabs screen, nothing is currently using it for the isMetric part of profile and the data isn't loaded until that screen. Would you prefer to keep lazy loading this data and have some checks for if it is already loaded or instead load profile right after login? How about refactoring the plates provider to use the profile provider instead of reading from shared prefs?

rolandgeider:
superseded by #1227

PR #1187, opened 2026-05-02
fix(routine): resolve in-place state mutation in GymStateNotifier and enable Routine immutability

pankaj-basnet:
In `offline-mode` branch, `Routine` model was being mutated in-place during exercise replacement, leading to UI sync issues where widgets do not rebuild despite the underlying data changing.

# Proposed Changes

* **Model (`lib/models/workouts/routine.dart`):**
    * Implemented `copyWith()` method for the `Routine` class.
    * Refactored `replaceExercise` from a `void` (mutating) method to a functional method that returns a new `Routine` instance.
    * Ensured nested lists (`sessions`, `dayData`, `dayDataGym`) are cloned using `List.from()` to prevent shared references between state snapshots.
* **Provider (`lib/providers/gym_state.dart`):**
    * Refactored `replaceExercises` in `GymStateNotifier` to capture the new `Routine` instance returned by the model.

* **Other refactor (`lib/widgets/core/image.dart`):**
    * Replace manual type checking with Dart switch expressions

### DIFF
--- a/lib/models/workouts/routine.dart
+++ b/lib/models/workouts/routine.dart
@@ -122,6 +122,38 @@ class Routine {
 
   Map<String, dynamic> toJson() => _$RoutineToJson(this);
 
+  Routine copyWith({
+    int? id,
+    DateTime? created,
+    String? name,
+    String? description,
+    bool? fitInWeek,
+    bool? isTemplate,
+    bool? isPublic,
+    DateTime? start,
+    DateTime? end,
+    List<Day>? days,
+    List<DayData>? dayData,
+    List<DayData>? dayDataGym,
+    List<WorkoutSession>? sessions,
+  }) {
+    return Routine(
+      id: id ?? this.id,
+      created: created ?? this.created,
+      name: name ?? this.name,
+      description: description ?? this.description,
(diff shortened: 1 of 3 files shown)


lib/models/workouts/routine.dart
```diff
@@ -237,8 +269,14 @@ class Routine {
     return groupedLogs;
   }
 
-  void replaceExercise(int oldExerciseId, Exercise newExercise) {
-    for (final session in sessions) {
+  Routine replaceExercise(int oldExerciseId, Exercise newExercise) {
+    final updatedRoutine = this.copyWith(
```
rolandgeider, line 273: As a side note, I think the fix only goes halfway since replaceExercise now returns a new Routine with new lists, but the elements inside (WorkoutSession, DayData, etc) are still the same instances as in the original routine so ideally we would push copies all the way down (Log.copyWith, SetConfigData.copyWith, if those even exist).

However this should fix the UI bug and I don't think we rely on the others, so this is fine. Another thing we could do is do another sweep on this by starting to make these fields all final, which is actually "correcter" (or at least looking at what doing that would entail, I suspect it will be a lot of small changes all over the codebase)

PR #1189, opened 2026-05-03
Implement server-side language validation on exercise description steps

### LINKED ISSUE #930: Add server side validation to exercise comments
### Use case

At the moment we validate that the language of the description of an exercise is
correct only when the complete exercise is submitted, which is awkward since the
user then has to jump back to the initial page. Ideally, we would like to 
validate it as soon as the description is edited (the user tries to navigate to
the next page).

See `Step3Description` and `Step4Translation`. There is a stub in
AddExerciseProvider, `validateLanguage` that sends the request to the backend.

### Proposal

_No response_

### Additional Context

_No response_

umaidshahid:
The `validateLanguage` stub in `AddExerciseProvider` was making the API call but ignoring the response and always returning false. This wires it up properly: the method now returns `null` on success or the server's error message on a language mismatch (caught from the 400 `WgerHttpException`). The stepper's `onStepContinue` is now async-aware for steps 2 (Description) and 3 (Translation) — it calls `validateLanguage` after form validation passes, blocks navigation, and displays the error inline if the check fails. Mocks regenerated to reflect the return type change from `Future<bool>` to `Future<String?>`.

# Proposed Changes
- Fix `validateLanguage` stub to parse the API response and return the server's error message on language mismatch
- Wire language validation into stepper navigation for the Description and Translation steps, blocking progress and showing the error inline
- Update mocks to reflect the `Future<bool>` → `Future<String?>` return type change

## Related Issue(s)

closes #930 


rolandgeider:
merged, thanks!

PR #1195, opened 2026-05-16
feat(settings): add in-app language switcher (#439)

### LINKED ISSUE #439: Feature request: Add an option to change language
## Use case

I can't be the only one who's phone isn't set to english but wants to see the app in english.

## Proposal

Should add an option in the settings to change the language used throughout the app. By default the app should use the same a system language.

PawelHaracz:
## Summary

Adds an in-app language picker in Settings, addressing #439. The user can override the system locale per-app; selecting **System default** clears the override and the existing `localeListResolutionCallback` resumes.

Also declares `CFBundleLocalizations` in `ios/Runner/Info.plist`, which enables the native iOS 16+ per-app language picker (Settings → wger → Language) as an alternative entry point on Apple platforms.

## Changes

- `lib/providers/user.dart` — `Locale? userLocale` field persisted via `SharedPreferencesAsync` (key `PREFS_USER_LOCALE`); mirrors the existing `themeMode` pattern. Stored value is matched back against `AppLocalizations.supportedLocales` on load to preserve script-bearing locales (e.g. `zh_Hant`, `pt_BR`).
- `lib/main.dart` — passes `locale: user.userLocale` to `MaterialApp`. When `null`, behavior is unchanged.
- `lib/widgets/core/settings/language.dart` *(new)* — `DropdownButton<Locale?>` listing all supported locales by their native names (`Polski`, `中文`, …) plus a *System default* entry, sorted alphabetically by display name.
- `lib/widgets/core/settings.dart` — wires `SettingsLanguage` into the existing Settings list, next to `SettingsTheme`.
- `lib/l10n/app_en.arb`, `lib/l10n/app_pl.arb` — adds `appLanguage` + `appLanguageSystem` strings. Other locales fall back to English via the standard delegate; happy to follow up if you'd prefer all 36 ARB files updated up front.
- `lib/helpers/consts.dart` — new `PREFS_USER_LOCALE` key.
- `ios/Runner/Info.plist` — declares `CFBundleLocalizations` for all 36 supported locales (enables iOS 16+ per-app language UI).

## Notes

- `intl` formatters already read `Localizations.localeOf(context)`, so dates/numbers should follow the new locale without further changes.

## Screenshots:
image
image

## Test plan

- [ ] CI green (`flutter test`)
- [X] Manual: switch to a non-default language → strings update without restart
- [X] Manual: switch back to *System default* → app follows OS locale again
- [X] Manual (iOS 16+): Settings → wger → Language shows native picker
- [X] Persistence: relaunch the app, last choice is restored

Closes #439.


rolandgeider:
merged, thanks!

PR #1203, opened 2026-05-23
Feature(#630): Show brand for ingredients

### LINKED ISSUE #630: App should show brand for ingredients
## Use case

Some products/ingredients are manufactured and sold under the same name but different brands, and also do not have the same nutritional values. It is therefore difficult to choose between equally named ingredients when searching for one.

## Proposal

The list obtained after a search in the nutrition tab should also display the brand, as fetched from eg. OpenFoodFacts. And the details panel too.

Maybe even, the search input could have a separate field for brand ? At least, search should also operate on brands : it is often quicker that way, because not so many brands will sell products with the same name but different nutritional values (at least in a given country).

More generally, inspiration, and maybe even some code, could be taken from [the Energize app](https://codeberg.org/epinez/Energize), which has way less features than wger but imho a better interface for food logging (including responsiveness).

anjakDev:
# Proposed Changes
As requested in #630, in the _Nutrition UI_, both the search results as well as the ingredient details show the `brand` of the item (if provided by the API). 
This makes it easier to quickly distinguish between multiple ingredient items with the same name, and find the right item to track. 

## Related Issue(s)

Closes #630 

### Screenshots
Simulator Screenshot - iPhone 17 - 2026-05-23 at 13 54 42
Simulator Screenshot - iPhone 17 - 2026-05-23 at 13 54 49


lib/widgets/nutrition/widgets.dart
```diff
@@ -212,8 +212,19 @@ class _IngredientTypeaheadState extends ConsumerState<IngredientTypeahead> {
                   : const CircleIconAvatar(
                       Icon(Icons.image, color: Colors.grey),
                     ),
-              title: Text(
-                ingredient.name,
+              title: Text.rich(
```
rolandgeider, line 215: ~~Is there a reason you added this in the title? Unless it breaks the UI (which I didn't test), I'd vote to simply show the brand in `subtitle`~~

Nevermind, we already have a subtitle

anjakDev: haha yeah, I saw the subtitle is used for the chips/tags, and I thought a while about what the best design choice is of where to put the brand, but adding it to the subtitle felt too crowded and also mismatched with the chips, and I didn't want to add a 'third' row as it would just increase the per-item space too much. Since _most_ items tend to have a short(er) name, I think `name` and `brand` sharing the title space looked okay. I'm open to suggestions though!

rolandgeider: no, you are right, the name and the brand belong more together.

If you want to fix a small thing, instead of changing the alpha here, use the color from one of the `Theme.of(context).colorScheme.xyz` (I think `onSurfaceVariant`?), then this also automatically works on the dark mode

anjakDev: Good point! I fixed it with the latest commit. I decided to still modify the alpha channel a little in the `ListTile` title, because otherwise I felt like there was too little contrast between the ingredient name and the brand (since they are sharing a line). For the `IngredientDetails` view, I use `onSurfaceVariant` unmodified, since it's positioned in the line below and has smaller fontSize anyway.

rolandgeider:
Merged 🚀

PR #1183, opened 2026-04-23
Offline mode

rolandgeider:
# Proposed Changes

Note: this needs a powersync service running additionally to the regular backend. The way this works is that you use a local sqlite database on the device and powersync takes care of making sure new entries are synced once you're connected to the internet again. Note that this doesn't make the whole application work offline, as some entries such as the routines need some backend logic that we don't want to replicate (also, this transparently and smoothly takes care of keeping the exercise db in sync). These parts are aware there is no network connectivity and won't endlessly try to fetch the data, so you can open your routine and go to the gym where there's no internet and log everything.

Also all providers were moved to riverpod since it made working with the new ones much easier and due to these refactors the PR should also get rid of all the weird loading bugs we had in the past.

## Related Issue(s)

Offline mode
Closes #6 🥳 
Closes #12

All kinds of miscellaneous "can't load x", "doesn't show y", "timeout at x" issues
Closes #51
Closes #548
Closes #558
Closes #577
Closes #722
Closes #861
Closes #867
Closes #881
Closes #936
Closes #983
Closes #1035
Closes #1054
Closes #1055
Closes #1063
Closes #1070
Closes #1089
Closes #1103
Closes #1114
Closes #1133 
Closes #1147
Closes #1152
Closes #1163
Closes #1193
Closes #1199

Since this also allows using all auth-options the backend does, also
Closes #302

Needs backend https://github.com/wger-project/wger/pull/1747

PR #1207, opened 2026-05-28
WIP: Default weight using profile

### LINKED ISSUE #1180: Use plate selection to imply default weight unit

jmccaull:
Alternative to https://github.com/wger-project/flutter/pull/1181. If this change set is preferred I will update the tests.

# Proposed Changes

- refactor main app to provide UserProvider as a dependency for RoutineProvider
- remove default weight unit from set config data
- adjust RoutineProvider.defaultWeightUnit to use UserProvider''s metric setting
- adjust RoutineProvider logic to utilize defaultWeightUnit when none provided

## Related Issue(s)

Closes https://github.com/wger-project/flutter/issues/1180

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -163,14 +163,15 @@ class MainApp extends StatelessWidget {
           update: (context, base, previous) =>
               previous ?? ExercisesProvider(WgerBaseProvider(base)),
         ),
-        ChangeNotifierProxyProvider2<AuthProvider, ExercisesProvider, RoutinesProvider>(
+        ChangeNotifierProxyProvider3<AuthProvider, ExercisesProvider, UserProvider, RoutinesProvider>(
           create: (context) => RoutinesProvider(
             WgerBaseProvider(Provider.of(context, listen: false)),
             Provider.of(context, listen: false),
+            Provider.of(context, listen: true),
             [],
           ),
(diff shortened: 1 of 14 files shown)


jmccaull:
The plate provider probably also needs to be refactored, but for that UserProvider would also need to first be converted to riverpod, correct? Do you want that addressed here?

rolandgeider:
>  but for that UserProvider would also need to first be converted to riverpod

We will get a riverpod userprovider once #1183 lands, so let's better wait a bit, no need to do that work twice

PR #1232, opened 2026-06-21
feat(auth): allow connecting to servers with self-signed certificates

### LINKED ISSUE #116: Allow connecting to self-hosted instances with self-signed certs
Currently, the app doesn't connect to my local instance because it has a self-signed cert, even though I've imported the CA in Android.

Can you add an option to allow logging with local cer (or an option to disable cert verification) when you select local server?

bhavit04:
## Use case

Closes #116.

Self-hosted wger instances served behind a self-signed (or otherwise invalid) TLS certificate currently can't be reached from the app — both API calls and image loading fail certificate validation, so users can't even view exercise images.

## Changes

- Adds an opt-in **"Allow self-signed certificates"** toggle to the advanced sign-in sheet, shown only when a **Self-hosted** server is selected.
- When enabled it installs an `HttpOverrides.global` (`WgerHttpOverrides`) that skips certificate validation. Because it's a global `dart:io` override, it transparently covers **both** the `http`-package API client *and* the `HttpClient` used by `NetworkImage` / `extended_image` for images — no per-call plumbing needed.
- The preference is stored in `AppSettings` (`PreferenceHelper`/`SharedPreferences`), re-applied on startup in `main()` *before* the first network request (auto-login probe), and applied immediately when toggled so the very next login already trusts the certificate.
- No-op on web, where `dart:io` networking / `HttpOverrides` doesn't apply (`kIsWeb` guard).

The toggle's subtitle warns that it disables TLS validation and should only be used for a trusted server, and it defaults to off.

## Known limitation

Native media players (the `video_player` plugin on Android/iOS) use their own networking stack and are **not** affected by `HttpOverrides`. Streaming exercise videos from a self-signed host would additionally need native config (Android `network_security_config.xml`, iOS ATS); that's intentionally left out of this PR to keep it focused on the connectivity + image path. Happy to follow up with a separate PR if wanted.

## Testing

- `flutter analyze` — no issues.
- New unit tests for the persisted setting (`test/core/settings_test.dart`) and the override helper (`test/helpers/http_overrides_test.dart`).
- New widget test verifying the toggle appears under *Self-hosted* and persists when tapped (`test/auth/auth_screen_test.dart`).
- Full suite green (the one failing `json_test.dart` timezone case also fails on a clean `master` and is unrelated).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rolandgeider:
that's probably a better (the only) solution in our case. I've moved this to the next version planning so it doesn't get lost 👍🏻

PR #1239, opened 2026-06-24
feat(weight): add time-range selector to body weight chart

### LINKED ISSUE #148: Allow zooming in the body weight chart
The body weight chart can potentially have a lot of entries that go back a long time, Having all of this historical data plotted is not always what the user wants and it would be a big improvement to be able to change this. Ideally we could add some kind of touch interface that allows the user to only show the charts in the selected part but adding three buttons ("all", "last year", "last 3 months") would be enough and is easier

bhavit04:
## Description

Closes #148

The body weight chart always plots the **entire** history. For users with a lot of entries going back a long time this gets cramped and hard to read, and there was no way to focus on a recent period.

This PR adds a segmented selector above the chart with three options — **All** / **Last year** / **Last 3 months** — exactly the approach suggested in the issue. Selecting a range restricts the plotted data (raw + 7-day average, and the per-plan/30-day sub-charts that derive from it).

Notes:
- **All** is the default, so existing behaviour is unchanged unless the user opts in — no regression risk for current users.
- The 7-day average is still computed over the full history and only filtered afterwards, consistent with how the existing per-plan and 30-day sub-charts are built.
- The change is scoped to the weight overview. The shared `getOverviewWidgetsSeries` helper only gains an **optional** `mainChartTitle` parameter so the header reflects the selected range; the measurements call site is untouched.
- New strings are added to `app_en.arb` only (`chartRange*` labels and `chartLastYearTitle` / `chartLast3MonthsTitle`); other locales fall back to English until translated.

## Testing

- Added a widget test (`test/weight/weight_screen_test.dart`) that verifies the selector renders, that restricting to "Last 3 months" hides out-of-range data, and that switching back to "All" restores the chart.
- `flutter analyze` clean; weight and measurement tests pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -534,6 +534,38 @@
       }
     }
   },
+  "chartLastYearTitle": "{name} last year",
+  "@chartLastYearTitle": {
+    "description": "last year chart of 'name' (e.g. 'weight', 'body fat' etc.)",
+    "type": "text",
+    "placeholders": {
+      "name": {
+        "type": "String"
+      }
+    }
+  },
+  "chartLast3MonthsTitle": "{name} last 3 months",
+  "@chartLast3MonthsTitle": {
+    "description": "last 3 months chart of 'name' (e.g. 'weight', 'body fat' etc.)",
+    "type": "text",
+    "placeholders": {
+      "name": {
+        "type": "String"
+      }
+    }
+  },
+  "chartRangeAll": "All",
+  "@chartRangeAll": {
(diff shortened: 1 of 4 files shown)


rolandgeider:
Thanks for the PR (and sorry for the wait)!

I've manually merged this branch into the [health-sync branch](https://github.com/wger-project/flutter/pull/1260) since we're reworking the way the charts are used, so I'm closing here, but your commits are not lost!

bhavit04:
Awesome, glad it was useful, thanks for folding it into the health-sync work!

PR #1255, opened 2026-07-07
Gym mode UI Updates

jontyms:
# Proposed Changes
Reworks gym mode around a single active-session screen: **everything lives on, or one tap away from, the log page.** The old per-set / per-timer chevron flow is replaced by a swipeable page-per-exercise with persistent chrome (session header + scrollable exercise queue).
This PR is heavily AI generated but I have tested and used this prototype for weeks. 

Screenshot_20260706_212532
Screenshot_20260706_212511
Screenshot_20260706_212445
Screenshot_20260706_214102


rolandgeider:
hey! thanks for the PR. I don't think I can merge this as it is (I think it makes the screen a bit restless), but there's definitely some good ideas here. I've marked this for the next release so it doesn't get lost

PR #1257, opened 2026-07-09
fix: make routineId nullable to prevent crash on free workout logs

### LINKED ISSUE #1250: Crash "Null check operator used on a null value" in WorkoutLogTableTable.map when a workout log has routine_id = NULL
### Priority/Impact

### Description

Opening a routine crashes the app with **"Null check operator used on a null value"** in `$WorkoutLogTableTable.map` (`database.g.dart`) if the user has any workout log whose `routine_id` is NULL.

**Root cause — schema mismatch between server and app:**

- The server explicitly allows logs without a routine (free/ad-hoc logging): `WorkoutLog.routine = models.ForeignKey('Routine', ..., null=True)` in `wger/manager/models/log.py`.
- The app's Drift table declares the column as **non-nullable** in `lib/database/powersync/tables/routines.dart`:
  ```dart
  IntColumn get routineId => integer().named('routine_id')();
  ```
  So the generated mapping does `data['routine_id']!` and throws as soon as PowerSync delivers such a row and any joined select over `manager_workoutlog` runs.

Related latent bug in the same code path: `Log` (`lib/features/routines/models/log.dart`) declares `late int routineId`, and `Log.fromSetConfigData` only assigns it inside `if (routineId != null) { ... }` — a null routine would leave the `late` field uninitialized (`LateInitializationError` on first read).

**Suggested fix:** make `routineId` nullable in the Drift table, the PowerSync schema and the `Log` model, matching the server model.

### Steps to reproduce

1. On the server (web UI or API), create a workout log that is not attached to a routine (`routine` = null) — e.g. free logging of an exercise.
2. In the app, let PowerSync sync.
3. Open any routine (or any screen that queries workout logs).
4. Crash:

```
Error Message: Null check operator used on a null value

#0  $WorkoutLogTableTable.map (package:wger/database/powersync/database.g.dart:5978)
#1  _SyncResultStructure.map (package:drift/.../select_with_join.dart:658)
#2  JoinedSelectStatement._mapResponse.<anonymous closure> (...)
```

Workaround for affected users (server side): attach the orphan logs to a routine, e.g.
`UPDATE manager_workoutlog SET routine_id = <id> WHERE routine_id IS NULL AND user_id = <uid>;`

### Server version

2.7.0a1 (self-hosted, docker compose, with PowerSync)

### Mobile app version

2.0.0 (iOS App Store)

mmustafasenoglu:
## Summary

Makes `routineId` nullable in `WorkoutLogTable` and `Log` model to fix a crash when workout logs have no associated routine.

## Problem

The server allows workout logs without a routine (free/ad-hoc logging): `routine = ForeignKey('Routine', ..., null=True)`. However, the app's Drift table declared `routineId` as **non-nullable**, causing:

```
Null check operator used on a null value
```

in `WorkoutLogTableTable.map` when PowerSync delivers a row with `routine_id = NULL`.

## Changes

- `WorkoutLogTable.routineId`: `integer()` → `integer().nullable()`
- `Log.routineId`: `late int` → `int?`
- `Log` constructor: `required this.routineId` → `this.routineId`
- `Log.toCompanion`: handle null routineId with `Value.absent()`

## Testing

Verified that the app no longer crashes when opening a routine that has workout logs without an associated routine.

Fixes #1250

### DIFF
--- a/lib/database/powersync/database.g.dart
+++ b/lib/database/powersync/database.g.dart
@@ -5973,7 +5973,7 @@ class $WorkoutLogTableTable extends WorkoutLogTable with TableInfo<$WorkoutLogTa
       routineId: attachedDatabase.typeMapping.read(
         DriftSqlType.int,
         data['${effectivePrefix}routine_id'],
-      )!,
+      ),
       sessionId: attachedDatabase.typeMapping.read(
         DriftSqlType.string,
         data['${effectivePrefix}session_id'],
@@ -6122,7 +6122,7 @@ class WorkoutLogTableCompanion extends UpdateCompanion<Log> {
   WorkoutLogTableCompanion copyWith({
     Value<String>? id,
     Value<int>? exerciseId,
-    Value<int>? routineId,
+    Value<int?>? routineId,
     Value<String?>? sessionId,
     Value<int?>? iteration,
     Value<int?>? slotEntryId,
--- a/lib/database/powersync/database.g.dart
+++ b/lib/database/powersync/database.g.dart
@@ -5685,9 +5685,9 @@ class $WorkoutLogTableTable extends WorkoutLogTable with TableInfo<$WorkoutLogTa
   late final GeneratedColumn<int> routineId = GeneratedColumn<int>(
     'routine_id',
     aliasedName,
-    false,
+    true,
     type: DriftSqlType.int,
-    requiredDuringInsert: true,
+    requiredDuringInsert: false,
   );
   static const VerificationMeta _sessionIdMeta = const VerificationMeta(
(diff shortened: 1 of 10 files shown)


rolandgeider:
thanks for the PR @mmustafasenoglu ! it seems you need to commit the changed generated files

mmustafasenoglu:
Hi @rolandgeider, thanks for the review! The generated files have been committed now — `database.g.dart` has been regenerated with the nullable `routineId` changes. The PR should be ready for another look.

rolandgeider:
merged, thanks!

PR #1276, opened 2026-07-23
Add elapsed workout timer to gym mode

### LINKED ISSUE #1130: Timer for current workout
### Use case

It would be nice to have a workout timer that tracks the time from pressing start until the last exercise is completed.

### Proposal

So you can see how long you have already spent training, and additionally automatically log the time you spend during the exercise instead of having to log it manually.

### Additional Context

_No response_

Arya-185:
# Proposed Changes

- Add a live elapsed-time display (`ElapsedWorkoutTimer`) to the gym mode navigation
  footer, next to the progress bar, so it is visible on every page of the workout
- The timer starts when the user taps "Start" on the first gym mode page; the same
  start time is used to pre-fill the workout session form, so the logged session
  duration matches what the timer showed
- Store the workout start with seconds precision (`DateTime workoutStart`) in
  `GymModeState`; the previous `TimeOfDay startTime` is kept as a getter so the
  session form behavior is unchanged
- Fix a bug where `GymStateNotifier.clear()` never actually reset the start time
  (passing `startTime: null` to `copyWith` was a no-op), which carried the previous
  workout's start time into the next workout in the same app session

## Related Issue(s)

Closes #1130

### DIFF
--- a/lib/features/routines/providers/gym_state.dart
+++ b/lib/features/routines/providers/gym_state.dart
@@ -170,7 +170,9 @@ class GymModeState {
   final List<PageEntry> pages;
   final int currentPage;
 
-  final TimeOfDay startTime;
+  /// Moment the workout was started, with full seconds precision. Set when
+  /// gym mode is opened and reset when the user taps "start".
+  final DateTime workoutStart;
   final DateTime validUntil;
 
   // User settings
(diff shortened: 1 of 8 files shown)


rolandgeider:
that one test seems to be flaky, we can ignore it

PR #1280, opened 2026-07-26
Don't lose a workout in gym mode when leaving it or when the app is killed

### LINKED ISSUE #1214: Save progress of workout
### Use case

If I am in the workout screen and doing my exercises it is not possible to leave the screen withoul losing all progress. The exercises are not checked anymore in the overview (the logs are still there).

Either we need an option to mark an exercise as done to get the workout to 100% done after leaving or the app should save te progress so we can go back and still all done exercises are still checked in the overview.

### Proposal

_No response_

### Additional Context

_No response_

andratwiro:
# Proposed Changes

Two related fixes for losing a workout in gym mode, as two independent commits.
The second one can be dropped if you'd rather discuss it separately.

**1. Ask before leaving a running workout**

Gym mode can be left with a single, easily accidental gesture. `GymModeScreen` has no
`PopScope`, so the iOS back swipe and the Android back gesture pop the route directly, and
the close button in `NavigationHeader` calls `pop()` unconditionally. Reaching for a music
or podcast app between two sets, or a slightly too generous swipe on the log page, ends the
workout with no way back into it.

A `PopScope` now blocks the pop while a workout is in progress, that is from the moment the
start page is left until the summary page is reached, and asks for confirmation instead.
The close button uses `maybePop()` so that it goes through the very same guard rather than
duplicating the policy. Dismissing the dialog without answering keeps the workout, so a
second stray tap can't leave it either. On the start page and on the summary nothing is in
progress, and closing stays a single tap as before.

**2. Continue a workout the OS interrupted**

The gym mode state only lived in memory. Riverpod keeps it for as long as the app runs, so
leaving gym mode and coming back already continued the workout (`initData` restores
`currentPage` for the same day within `DEFAULT_DURATION`), but an app that the OS kills in
the background loses the session, the elapsed time and which sets had been done. The saved
logs themselves survive, since `LogPage` writes each set on save.

The progress is now also written to the preferences and read back when gym mode is opened.
Only what can't be recalculated is stored, the pages still come from the routine, which is
why the finished log pages are referenced by page index instead of by their UUID (those are
generated anew on every `calculatePages()`).

The stored progress is deliberately conservative. It is only used for the same day and
iteration, only while the workout is plausibly still running (the existing 5 hour window),
and only if the routine still has the page it stopped on. A workout still in memory wins
over the stored one, and finishing a workout clears it.

There is a product decision hidden in here that you may want to make differently: how long
a killed workout should stay resumable. This reuses the 5 hour `validUntil` that gym mode
already applies to its in-memory state, so behaviour stays consistent, but it does mean a
workout can be picked up again hours later after a phone restart.

## Two things worth flagging

**The tests in `screens/gym_mode_test.dart` don't isolate their preferences.** Installing a
fresh `InMemorySharedPreferencesAsync` in `setUp` does not stop what one test writes from
being visible in the next. It has stayed invisible so far only because the test that sets
`PREFS_SHOW_EXERCISES` is the last one in the file; adding tests after it makes them fail.
The tests here set the preferences they depend on explicitly instead of relying on the
reset. Happy to fix the fixture itself in a separate PR if you'd like it done properly.

**Where I'd like to take this next**, and the reason for the Discord thread rather than
more code: Strong never lets you lose a workout because leaving gym mode *minimises* it
instead of exiting, with a persistent bar and an ongoing notification bringing you back,
and only an explicit "Finish" ends it. If you'd rather go that way, this PR's dialog is
what would be replaced, while commit 2 is a prerequisite for it either way.

## Related Issue(s)

See also #1255, which reworks the same screen and would need one of the two rebasing.

Tested with Flutter 3.44.4, as pinned in `.github/actions/flutter-common`:
`flutter test test/features/routines` is green (245 tests), `flutter analyze lib test`
reports no issues.

### DIFF
--- a/lib/features/routines/models/workout_progress.dart
+++ b/lib/features/routines/models/workout_progress.dart
@@ -1,0 +1,72 @@
+/*
+ * This file is part of wger Workout Manager <https://github.com/wger-project>.
+ * Copyright (c) 2026 wger Team
+ *
+ * wger Workout Manager is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/// The progress through a running workout, as stored in the preferences
+///
+/// Only the parts that can't be recalculated are kept. The pages themselves are
+/// derived from the routine again on the next start, which is also why the
+/// finished log pages are referenced by their page index: the UUIDs are
+/// generated anew every time the pages are calculated.
+class WorkoutProgress {
+  final int dayId;
+  final int iteration;
+  final int currentPage;
+  final DateTime workoutStart;
+  final DateTime validUntil;
+
+  /// Absolute page indices of the log pages already marked as done
+  final List<int> donePageIndices;
+
+  const WorkoutProgress({
+    required this.dayId,
+    required this.iteration,
+    required this.currentPage,
+    required this.workoutStart,
+    required this.validUntil,
+    required this.donePageIndices,
+  });
+
+  factory WorkoutProgress.fromJson(Map<String, dynamic> json) => WorkoutProgress(
+    dayId: json['dayId'] as int,
+    iteration: json['iteration'] as int,
+    currentPage: json['currentPage'] as int,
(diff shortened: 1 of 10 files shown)


bentucker:
FWIW I have been running an equivalent of commit 2 on my own build since
June (self-hosted, 6 workouts a week), and the conservative restore rules
here match what turned out to matter in practice: same day and iteration
only, bounded by the existing 5 hour window, in-memory state winning over
the stored copy, and clearing on finish. The page-index reference is also
the right call, I initially tried stable page ids and it was not worth it.

One gap worth closing while you are in here: PREFS_WORKOUT_PROGRESS is only
removed when a workout finishes. Logout and the different-user wipe don't
touch it, so on a shared device the next account can walk into gym mode on
the right day and inherit the previous user's position. In my build I clear
the pointer in the same place the local data wipe runs, which covers both
full logout and the user-switch wipe; a one-line remove there closes it.

Two things I could offer as follow-ups once this lands, to keep this PR
small: a dashboard "resume workout" card that deep links back into gym mode
(discoverability was the main complaint from my own use, reopening the day
is easy to miss), and a fairly large gym_state test suite for the
persistence rules that should port over with little work.

PR #1287, opened 2026-07-27
Fix crash when copying a log with no hydrated exercise

### LINKED ISSUE #1286: LateInitializationError: Field 'exerciseObj' has not been initialized when copying a past log in gym mode
### Priority/Impact

### Description

In gym mode, tapping an entry in the "Workout logs" list (the past logs with the copy icon) on an exercise's log page shows an error dialog instead of copying the values into the form.

### Steps to reproduce
1. Start a workout in gym mode
2. Go to the log page of an exercise that already has previous logs
3. Tap one of the entries in the "Workout logs" list (the ones with the copy icon)

### Expected behaviour

The reps and weight of the tapped log are copied into the form and a "data copied" snackbar is shown.

### Actual behaviour

An error dialog appears and nothing is copied.

```
LateInitializationError: Field 'exerciseObj' has not been initialized.
#0   Log.copyWith (package:wger/features/routines/models/log.dart)
#1   GymLogNotifier.setLog (package:wger/features/routines/providers/gym_log_notifier.dart:41)
#2   LogsPastLogsWidget.build.<anonymous closure>.<anonymous closure>
     (package:wger/features/routines/widgets/gym_mode/log_page.dart:264)
#3   _InkResponseState.handleTap (package:flutter/src/material/ink_well.dart:1224)
```

It happens consistently: the past logs on that page are read from the local database and only carry an exercise ID, and copying one reads an exercise that was never set.

Image

### Server version

_No response_

### Mobile app version

Play store, 2.0.2

MatzHilven:
# Proposed Changes

- Fix `LateInitializationError: Field 'exerciseObj' has not been initialized` when copying a
  past log in gym mode. `Log.exerciseObj` becomes a nullable backing field with a throwing
  getter plus an `exerciseObjOrNull` accessor, and `Log.copyWith` only carries the exercise
  over when the source log actually has one.
- `GymLogNotifier.setLog` takes an optional `exercise`, and `LogsPastLogsWidget` passes the
  one it already has, so gym-mode state never ends up holding a log without an exercise.
- Make the gym-mode log fixtures match what drift really returns, so the regression is
  actually covered.

## Related Issue(s)

Closes #1286

## Details

Tapping an entry in the "Workout logs" list on an exercise's log page threw:

```
LateInitializationError: Field 'exerciseObj' has not been initialized.
#0   Log.copyWith (package:wger/features/routines/models/log.dart)
#1   GymLogNotifier.setLog (package:wger/features/routines/providers/gym_log_notifier.dart:41)
#2   LogsPastLogsWidget.build.<anonymous closure>.<anonymous closure>
     (package:wger/features/routines/widgets/gym_mode/log_page.dart:264)
```

### Two different ways a Log gets built

`Log.exerciseObj` was declared `late Exercise`, and only one code path ever assigned it:
`routines_notifier.dart:153`, which hydrates the logs hanging off a routine's sessions.

The gym-mode past-logs list does not use that path. It watches `pastExerciseLogsProvider`,
which goes to `WorkoutLogRepository.watchLogsByExerciseDrift()`. That query left-outer-joins
the repetition and weight unit tables and hydrates `repetitionUnit` / `weightUnit` on each
row, but never touches the exercise. Since `WorkoutLogTable` is `@UseRowClass(Log)`, drift
builds every row through the constructor, which only sets `exerciseId`.

So logs from the DB legitimately have no `exerciseObj` — and `copyWith` ended with an
unconditional read of it (`out.exerciseObj = exerciseObj;`). That is a precondition the real
data path simply cannot satisfy, and tapping a past log is exactly the case that hits it.

Deliberately *not* fixed by hydrating the exercise inside `watchLogsByExerciseDrift`:
exercises carry translations, images, muscles and equipment, and joining all of that per log
would be a real performance regression for a value most callers don't need. `Log.toCompanion()`
only persists `exerciseId`, so an unhydrated exercise never affects what is written to the DB.

### Why CI did not catch it

Both existing fixtures hid the bug:

- `log_page_test.dart` mocked `watchLogsByExerciseDrift` with
  `getTestRoutine().filterLogsByExercise(...)`, whose logs *are* hydrated — unlike real drift
  output. The mock is now mapped through a helper that strips the exercise, so the
  "copy from past log" test is a genuine regression test.
- `gym_log_notifier_test.dart` worked around it explicitly, with a comment saying
  `// Log.copyWith reads exerciseObj internally, so it must be set before setLog ... is invoked`
  followed by `..exerciseObj = testBenchPress`. That workaround was the bug, written down.
  It is gone.

### Note

`SlotEntry` has the same `late Exercise exerciseObj` pattern, but its hydration path
(`routines_notifier.dart:165`) is sound, so it is left alone to keep this PR focused.

### DIFF
--- a/lib/features/routines/models/log.dart
+++ b/lib/features/routines/models/log.dart
@@ -43,7 +43,25 @@ class Log {
   String? id;
 
   late int exerciseId;
-  late Exercise exerciseObj;
+
+  Exercise? _exerciseObj;
+
+  /// The exercise this log belongs to, throws if it was never hydrated
+  ///
+  /// Logs read from the database only carry [exerciseId], joining the whole
+  /// exercise with its translations, images and muscles would be wasteful for
+  /// a value most callers don't need. Use [exerciseObjOrNull] for those.
+  Exercise get exerciseObj {
+    if (_exerciseObj == null) {
+      throw StateError('Log $id has no hydrated exercise (exercise ID $exerciseId)');
+    }
+    return _exerciseObj!;
+  }
+
+  set exerciseObj(Exercise value) => _exerciseObj = value;
+
+  /// Like [exerciseObj] but returns null instead of throwing
+  Exercise? get exerciseObjOrNull => _exerciseObj;
 
   int? routineId;
   String? sessionId;
@@ -153,7 +171,11 @@ class Log {
       out.weightUnitObj = weightUnitObj;
       out.weightUnitId = weightUnitObj.id;
     }
-    out.exerciseObj = exerciseObj;
+
+    // Logs read from the database have no exercise, so only copy an existing one
+    if (_exerciseObj != null) {
+      out.exerciseObj = _exerciseObj!;
+    }
 
     return out;
   }
--- a/lib/features/routines/providers/gym_log_notifier.dart
+++ b/lib/features/routines/providers/gym_log_notifier.dart
@@ -19,6 +19,7 @@
 import 'package:clock/clock.dart';
 import 'package:logging/logging.dart';
 import 'package:riverpod_annotation/riverpod_annotation.dart';
+import 'package:wger/features/exercises/models/exercise.dart';
 import 'package:wger/features/routines/models/log.dart';
 import 'package:wger/features/routines/models/repetition_unit.dart';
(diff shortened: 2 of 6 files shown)


rolandgeider:
merged, thanks!

PR #1290, opened 2026-07-28
Gym mode: add optional sticky set values

BrianHotopp:
# Proposed Changes

New **"sticky set values"** gym-mode setting (default off): the next set of an exercise pre-fills from the values you last logged — weight, reps, RiR, units — instead of the planned ones.

Use case: dialing in the weight over the first sets of an exercise. Today every set re-fills from the plan, so the adjustment has to be re-entered each time.

- Only values are copied: id, session, slot entry and targets stay a fresh template, so saving still creates a correct new log row.
- Per exercise, cleared at workout end. No behavior change unless enabled.
- Client-side only, no backend changes.

## Related Issue(s)

- Independent of #1289; both touch `gym_state*.dart`, so a small rebase may be needed depending on merge order — happy to do it.

`flutter analyze` clean; `flutter test test/features/routines` 244/244 (+6 tests).

### DIFF
--- a/lib/features/routines/providers/gym_state.dart
+++ b/lib/features/routines/providers/gym_state.dart
@@ -21,6 +21,7 @@ import 'package:flutter/material.dart';
 import 'package:wger/core/uuid.dart';
 import 'package:wger/features/exercises/models/exercise.dart';
 import 'package:wger/features/routines/models/day_data.dart';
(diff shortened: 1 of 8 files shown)


BrianHotopp:
Should this be default true?

rolandgeider:
I'd say false is better

PR #1313, opened 2026-08-12
fix(auth): do not await PowerSync disconnect during session clear

### LINKED ISSUE #1312: Rejected token refresh deadlocks the app: every request times out until re-login is impossible
### Priority/Impact

High (critical issue, blocks workflow)

### Description

When the server rejects the app's refresh token (HTTP 400 on
/allauth/app/v1/tokens/refresh, e.g. because the session behind it expired),
the app is supposed to clear the session and show the login screen. Instead it
deadlocks: every subsequent API request hangs and times out after 15 seconds,
on every network, and restarting the app does not help. The user is never
shown the session-expired message and the app looks completely broken (no
routines, no logging, no sync) even though the server and the network are
fine.

The cycle:

1. The failed refresh runs clearSessionOnly() inside the single-flight
   refresh future (_refreshInFlight in AuthNotifier).
2. _resetSession awaits _disconnectPowerSyncIfBuilt() on the keep-data path.
3. PowerSync's disconnect() waits for its in-flight sync fetch to abort.
4. That sync fetch goes through AuthHttpClient, whose pre-emptive refresh is
   awaiting the very same _refreshInFlight future.

refresh -> disconnect -> sync fetch -> refresh: a three-party circular wait.
The refresh future never completes, and because it is single-flight, every
later request that needs a token refresh queues behind it forever. Restarting
the app re-triggers the same rejected refresh during startup revalidation, so
the deadlock re-forms immediately.

Diagnosed in production on a self-hosted instance: all endpoints (including
trivial ones like /api/v2/trophy/) timed out from the app for days, while
curl from the same device, even run as the app's own UID, got sub-second
responses. Logcat showed the refresh 400 followed by "clearing session,
keeping local DB" and then no further refresh attempts, only endless
15-second timeouts.

The fix is small: on the keep-data path there is no DB wipe to guard, so the
PowerSync disconnect does not need to be awaited before publishing the
cleared state. Making it fire-and-forget breaks the cycle: the refresh future
completes, the cleared state is published, the login screen (with the
session-expired message) appears, and the wedged sync fetch resumes with no
credential, gets a 401 and PowerSync backs off. The wipe path (explicit
logout) keeps its current ordering.

I will open a PR with the fix and a regression test.

### Steps to Reproduce

1. Log in and let PowerSync connect (local-first mode).
2. Invalidate the session server-side so the next token refresh returns 400
   (e.g. delete the session/refresh token, or wait for it to expire).
3. Use the app while a sync iteration is in flight: the pre-emptive refresh
   fires, gets the 400 and starts the session teardown.
4. Observe: every fetch now logs "TimeoutException after 0:00:15" retries,
   the session-expired message never appears, and force-stopping and
   restarting the app reproduces the hang immediately.

### Server version

2.7.0a1 (self-hosted, docker)

### Mobile app version

master (also reproduced on a 2.0.x-based build)

bentucker:
# Proposed Changes

- Stop awaiting the PowerSync disconnect in the keep-data session reset
  (clearSessionOnly). The await closes a three-party circular wait when a
  refresh token is rejected: the refresh future runs the session teardown,
  the teardown awaits PowerSync's disconnect, the disconnect waits for the
  in-flight sync fetch to abort, and that fetch is awaiting the same
  single-flight refresh future through AuthHttpClient's pre-emptive refresh.
  The refresh future then never completes and every later authenticated
  request queues behind it forever (15 second timeouts on every endpoint,
  surviving app restarts). See the issue for the full production diagnosis.
- The disconnect is now fire-and-forget on this path. The DB is deliberately
  kept, so there is no wipe to race with: once the cleared state is
  published, the wedged sync fetch resumes without a credential, gets a 401
  and PowerSync backs off, letting the deferred disconnect finish. The wipe
  path (explicit logout) keeps its wipe-before-publish ordering.
- Make the disconnect hook injectable (visibleForTesting field) so the
  regression test can install a never-completing disconnect;
  PowerSyncDatabase is a base class and cannot be faked directly.
- Add a regression test: a rejected refresh must complete and publish the
  logged-out state even when the PowerSync disconnect hangs forever.

## Related Issue(s)

Closes #1312

## Checklist

- [x] Tests pass locally (flutter test, 1235 tests)
- [x] flutter analyze reports no issues on the changed files
- [x] New behaviour is covered by a regression test

### DIFF
--- a/lib/core/network/auth_notifier.dart
+++ b/lib/core/network/auth_notifier.dart
@@ -94,6 +94,12 @@ class AuthNotifier extends _$AuthNotifier {
   @visibleForTesting
   int userSwitchWipeCount = 0;
 
+  /// Stops the PowerSync sync loop on the keep-data session-reset path.
+  /// Injectable because `PowerSyncDatabase` is a `base` class and cannot be
+  /// faked in tests.
+  @visibleForTesting
+  Future<void> Function() disconnectPowerSync = _disconnectBuiltPowerSync;
+
   @override
   Future<AuthState> build() async {
     _client = ref.read(authHttpClientProvider);
@@ -906,9 +912,10 @@ class AuthNotifier extends _$AuthNotifier {
   /// [logout] instead, which performs a full wipe.
   Future<void> clearSessionOnly() => _resetSession(wipeLocalData: false, sessionExpired: true);
 
-  /// Shared body for [logout] and [clearSessionOnly]. PowerSync is touched
-  /// before the state mutation so a reader observing the post-reset state
-  /// can never race ahead and re-attach to a DB we're about to wipe.
+  /// Shared body for [logout] and [clearSessionOnly]. On the wipe path
+  /// PowerSync is touched before the state mutation so a reader observing the
+  /// post-reset state can never race ahead and re-attach to a DB we're about
+  /// to wipe.
   ///
   /// [sessionExpired] marks the reset as involuntary in the published state,
   /// so the login screen can tell the user why they were logged out.
@@ -927,7 +934,11 @@ class AuthNotifier extends _$AuthNotifier {
         wiped = false;
       }
     } else {
-      await _disconnectPowerSyncIfBuilt();
+      // Deliberately not awaited: this path runs inside the single-flight
+      // refresh future, and the disconnect can block on a sync fetch that is
+      // itself awaiting that future (refresh -> disconnect -> sync fetch ->
+      // refresh deadlock). The DB is kept, so there is no wipe to race with.
+      unawaited(disconnectPowerSync());
     }
 
     state = AsyncData(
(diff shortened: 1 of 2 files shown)


lib/core/network/auth_notifier.dart
```diff
@@ -927,7 +934,18 @@ class AuthNotifier extends _$AuthNotifier {
...
+      // disconnect can block on an in-flight sync fetch that is itself
+      // awaiting that same refresh future through AuthHttpClient's
+      // pre-emptive refresh: refresh -> disconnect -> sync fetch -> refresh.
+      // Awaiting here closes that cycle and permanently wedges every
+      // authenticated request behind the never-completing refresh
+      // (observed in production: all fetches timing out after the server
+      // rejected a refresh token). The DB is kept on this path, so there is
+      // no wipe to race with; the disconnect is only a courtesy stop of the
+      // sync loop and can safely finish after the state is cleared.
+      unawaited(disconnectPowerSync());
```
rolandgeider, line 941: could you trim a bit this comment, then we can merge this :)

bentucker: Done, trimmed it down (and the related doc comments too).

PR #1322, opened 2026-08-16
fix(core): make UI side effects and provider reads safe during build

### LINKED ISSUE #1321: UI side effects and provider reads during widget build: error dialogs wedge shut, sync dialog and startup crash
### Priority/Impact

High (critical issue, blocks workflow)

### Description

Several code paths perform UI side effects or provider reads while the widget
tree is building. They look harmless and pass review, but they crash or wedge
the app whenever they run in the wrong frame phase, which in practice happens
around auth token refreshes and connectivity flaps. I hit two separate
production incidents from this bug class on my self-hosted setup within one
week; diagnosing them turned up more instances of the same pattern that had
simply not fired yet.

The class has three shapes:

1. UI side effects from error handlers. FlutterError.onError routes to
   showGeneralErrorDialog, which calls showDialog synchronously. When the
   original error happens during build (exactly the case onError exists for),
   the route push trips the navigator's !_debugLocked assertion. Worse, the
   one-dialog guard flag is set before showDialog throws and its release is
   attached to a future that never materialises, so the flag wedges shut:
   every later error dialog is silently suppressed while nothing is on
   screen. The app appears blank and dead until a restart, with logcat full
   of "Suppressing error dialog, one is already visible". The snackbar
   helpers have the same problem via the messenger's setState ("setState()
   or markNeedsBuild() called during build").

2. ref.read inside route builders. The sync status dialog's builder reads
   wgerBaseProvider and networkStatusProvider; the barcode scan dialog's
   builder reads ingredientRepositoryProvider (and creates its search future
   inline, so it also re-fires the search on every dialog rebuild). Route
   builders run during build; reading a provider whose dependency chain is
   dirty (wgerBase -> auth goes dirty on every token refresh) flushes it
   there, and Riverpod schedules a provider scope refresh mid-build:
   "setState() or markNeedsBuild() called during build", caught by Riverpod
   as EXCEPTION CAUGHT BY RIVERPOD. Reproduced in production by tapping the
   sync icon right after a token refresh while sync was stalled.

3. Synchronous provider reads in provider create bodies that widgets watch
   during build. routineHydrationProvider is created by a ref.watch in the
   dashboard's build; its create chain read routinesRepositoryProvider
   synchronously and crashed on a cold start right after a token refresh, on
   an untouched screen. TrophyStateNotifier.build() has the same shape
   (ref.read(networkStatusProvider) with no preceding yield), reachable in
   the post-login window because _invalidatePostLoginProviders invalidates
   the trophy provider before networkStatusProvider, so the first trophy
   watcher rebuilds against a dirty leaf.

I will open a PR that fixes all known instances and adds two small shared
helpers (lib/core/build_safety.dart) so the safe form is the easy form:
runAfterFrame() for deferred UI side effects and yieldPastBuild() for
provider create bodies. The error dialog guard also becomes wedge-proof
(released when the context is gone or the push fails), the snackbar helpers
re-resolve the messenger inside the callback (the captured one can be
disposed during auth teardown), the dialog builders snapshot their providers
in the tap handler, and post-login invalidation runs leaf-first. Regression
tests cover each shape, including an "invoke it during build" harness for
the error dialog and snackbar paths.

A possible follow-up, if there is interest: these three shapes are
mechanical enough to encode as custom_lint rules so new instances cannot be
written; happy to discuss that separately.

### Steps to Reproduce

For shape 1 (the most user-visible):

1. Run a debug build connected to a self-hosted server.
2. Cause any widget error during build while an error dialog would be shown
   (or simply let a provider read crash as in shape 2/3 below).
3. Observe the !_debugLocked assertion, then "Suppressing error dialog, one
   is already visible" on every subsequent error, with a blank unresponsive
   UI until force-stop.

For shape 2:

1. Let the access token approach expiry (or invalidate the session server
   side) so the next request triggers a refresh, dirtying the wgerBase
   provider chain.
2. Tap the sync status (cloud) icon in the app bar before anything else
   rebuilds.
3. Observe EXCEPTION CAUGHT BY RIVERPOD with "setState() or markNeedsBuild()
   called during build" pointing at the dialog builder's ref.read.

For shape 3:

1. Cold start the app with a valid session whose access token is within the
   refresh leeway, so startup triggers a token refresh.
2. The dashboard's routine hydration watch crashes during build with the
   same markNeedsBuild error, with no user interaction at all.

### Server version

2.7.0a1 (self-hosted, docker)

### Mobile app version

master (reproduced on 2.0.4 plus master as of 6eb61979)

bentucker:
# Proposed Changes

- Add lib/core/build_safety.dart with the two idioms this PR applies
  everywhere, so the safe form is the easy form: runAfterFrame() defers UI
  side effects (dialogs, snackbars) past the build phase, and
  yieldPastBuild() yields provider create bodies past their synchronous
  phase. The yield is a microtask, not a zero timer, because widget tests
  assert !timersPending at teardown.
- Defer the error dialogs and both error snackbars to the next frame.
  Error handlers can run during build, where pushing a route asserts
  (!_debugLocked) and the messenger's setState throws. The one-dialog guard
  is now wedge-proof: it is released when the dialog closes, when the
  context is gone by the time the frame fires (logged), or when the push
  itself fails. Previously the secondary exception left the guard stuck and
  every later error dialog was silently suppressed while the app sat blank
  and unresponsive. The snackbar helpers re-resolve the messenger inside
  the callback, since the captured one can be disposed during auth
  teardown.
- Snapshot providers in the tap handler instead of the route builder for
  the sync status dialog and the barcode scan dialog. Route builders run
  during build, and ref.read on a dirty provider chain (any token refresh
  dirties wgerBase -> auth) flushes it there, scheduling a provider scope
  refresh mid-build. The barcode change also stops the search future from
  being re-created on every dialog rebuild.
- Yield before provider work in the routine hydration chain
  (routineHydrationProvider is created by a ref.watch during the
  dashboard's build) and move TrophyStateNotifier's reachability read into
  its existing microtask. Run the post-login invalidations leaf-first so
  dependents never rebuild against a dirty networkStatusProvider.
- Regression tests for each shape: an error dialog requested during build
  appears on the next frame without asserting and the guard is released
  afterwards; a snackbar requested during build appears without asserting;
  the sync icon opens its dialog with tap-time snapshots; and the hydration
  provider performs no provider reads during its synchronous create phase
  (this last test fails against the previous code).

All three shapes were hit in production on a self-hosted setup within one
week (details and logs in the issue). A possible follow-up is encoding the
three patterns as custom_lint rules so new instances cannot be written;
happy to discuss whether that is wanted.

## Related Issue(s)

Closes #1321

Tested with the pinned Flutter: flutter test is green (1240 tests, 5 new),
flutter analyze lib test reports no issues.

### DIFF
--- a/lib/core/build_safety.dart
+++ b/lib/core/build_safety.dart
@@ -1,0 +1,33 @@
+/*
+ * This file is part of wger Workout Manager <https://github.com/wger-project>.
+ * Copyright (c) 2026 wger Team
+ *
+ * wger Workout Manager is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import 'package:flutter/widgets.dart';
+
+/// Runs [action] on the next frame: UI side effects (dialogs, snackbars,
+/// navigation) are illegal during build, but error handlers and provider
+/// listeners can fire there. Schedules a frame so idle-time requests run too.
+void runAfterFrame(VoidCallback action) {
+  WidgetsBinding.instance.addPostFrameCallback((_) => action());
+  WidgetsBinding.instance.scheduleFrame();
+}
+
+/// Yields past the synchronous phase of a provider create or widget
+/// life-cycle callback: reading a dirty provider there flushes it mid-build
+/// and crashes. A microtask, not a timer, since widget tests assert
+/// !timersPending at teardown.
+Future<void> yieldPastBuild() => Future<void>.microtask(() {});
--- a/lib/core/error_dialogs.dart
+++ b/lib/core/error_dialogs.dart
@@ -23,6 +23,7 @@ import 'package:flutter_html/flutter_html.dart';
 import 'package:json_annotation/json_annotation.dart';
 import 'package:logging/logging.dart';
 import 'package:url_launcher/url_launcher.dart';
+import 'package:wger/core/build_safety.dart';
 import 'package:wger/core/errors.dart';
 import 'package:wger/core/exceptions/http_exception.dart';
 import 'package:wger/core/keys.dart';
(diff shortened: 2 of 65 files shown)


rolandgeider:
merged, thanks

PR #1326, opened 2026-08-20
Handle duplicate muscles on muscle overview

### LINKED ISSUE #145: Handle duplicate muscles on muscle overview
Currently `MuscleWidget` simply renders the list of main and secondary muscles given to it. When a muscle is both main and secondary, only the main background should be drawn (this is something that, if the exercises are saved sensibly should not happen, but we still need to account for it)

Screenshot_1648498312

charugupta-dev:
## Description
This PR resolves #145 by ensuring that when a muscle is categorized as both a primary and secondary muscle for an exercise, only the primary muscle styling is rendered in `MuscleWidget`.

### Changes Made
- In `MuscleWidget` initialization, filtered `musclesSecondary` to exclude any muscle whose `id` already exists in the primary `muscles` list.
- Deduplicated muscles by `id` within both `muscles` and `musclesSecondary` to prevent redundant SVG layers in the rendering stack.
- Added comprehensive unit and widget tests in `test/features/exercises/widgets/muscle_widget_test.dart` covering:
  - Front and back orientation filtering.
  - Primary muscle precedence when a muscle is passed in both lists.
  - Deduplication of duplicate entries within the same list.

Fixes #145

### DIFF
--- a/lib/features/exercises/widgets/exercises.dart
+++ b/lib/features/exercises/widgets/exercises.dart
@@ -138,23 +138,10 @@ class MuscleWidget extends StatelessWidget {
     List<Muscle> musclesSecondary = const [],
     this.isFront = true,
   }) {
-    final frontMuscles = muscles.where((m) => m.isFront == isFront);
-    final mainMuscleIds = <int>{};
-    this.muscles = [];
-    for (final m in frontMuscles) {
-      if (mainMuscleIds.add(m.id)) {
-        this.muscles.add(m);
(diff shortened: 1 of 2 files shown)


rolandgeider:
Hi, thanks for the PR, it seems this was already fixed in master, but the issue wasn't closed. We can compare the muscles by values, that works as well here, no need to use the IDs. But we can definitely keep the tests which we didn't have before

